PR #45, opened 2024-01-25
Support for NeuropixelsV2 public and beta headstages

### LINKED ISSUE #37: NeuropixelsV2 configuration and streaming

### LINKED ISSUE #48: Runtime exception "SafeHandle is closed" when stopping Nueorpixels2.0Beta Acqusition
This issue pertains to 69c0d35. When I run a workflow for Neuorpixels2.0Beta acquisition, everything works great. However, when I press stop I get an Runtime Error about a SafeHandle being closed:

image

Then the next time I try to run, we get our old friend:

image

glopesdev:
Leveraging the passthrough serializer interface, this PR bootstraps the required infrastructure to configure and stream data from NeuropixelsV2 devices. The current implementation has been tested on a V2 beta headstage but will require validation and testing with other probes and public devices.

The device configuration call has also been moved inside the `ContextTask` class to ensure configuration is disposed in the correct order in both exceptional termination and cancellation.

Fixes #37 
Fixes #48

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2BetaData.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2BetaData.cs
@@ -15,6 +15,8 @@ namespace OpenEphys.Onix
 
         public int BufferSize { get; set; } = 30;
 
+        public NeuropixelsV2Probe ProbeIndex { get; set; }
+
         public unsafe override IObservable<NeuropixelsV2BetaDataFrame> Generate()
         {
             var bufferSize = BufferSize;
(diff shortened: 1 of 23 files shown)


OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2BetaData.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Linq;
+using Bonsai;
+using OpenCV.Net;
+
+namespace OpenEphys.Onix
+{
+    public class NeuropixelsV2BetaData : Source<NeuropixelsV2BetaDataFrame>
```
jonnew, line 11: I think I would be happier if `NeuropixelsV2BetaDataFrame` did not include an `ProbeIndex` field and instead, this was somehow made part of this Class's Device name: 

image

Then instead of generating buffers for all probe Indices, only the selected ProbeIndex would be used and the field could be removed from the `NeuropixelsV2BetaDataFrame`

jonnew: The same idea could be applied to non-beta-test variant as well.

jonnew (commented):
Aside from these comments see this:

https://github.com/neurogears/onix-refactor/issues/47
https://github.com/neurogears/onix-refactor/issues/48

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2BetaData.cs
glopesdev: @jonnew I am not keen to have probe index in the device name property as this would require individual probes to be registered in the device table (virtual or actual).

However, this feedback did made me think we might want to return the probe SN instead, as it could be used as a unique ID, e.g. when merging the output of multiple ONIX contexts together.

Is the worry here that people will simply record the data of both probes into a single file without splitting it? This is a valid concern, and in this case I have a few alternative ideas we used in other packages which might solve this.

jonnew: That is the worry. When you have a node that says NeuorpixelsData, its a pretty reasonable assumption that it should be data from a neuropixel rather than a stream without guarantee of order coming from two neuropixels at once that needs to be sorted post hoc to be meaningfully plotted or recorded. Especially considering that there will be another node in the workflow called something like BNOData that produces data from a single IMU.

jonnew: To be concrete about it this was the workflow I threw together when testing my ill-faited BNO node. 

image

I was looking at `AmplifierData` going why does it looks so weird? The issue is that you can expose fields from the frame without regard to each other. The fact is that `AmplifierData` is inextricably linked to `ProbeIndex`, but this cannot be enforced, nor should it be, in a workflow.

glopesdev:
@jonnew One thing that maybe we didn't finalize yet is how to deal with the gain conversions, and also the logic for loading the corrections from file, I will try to make a prototype of this and incorporate in this branch.

glopesdev:
@jonnew Actually, I think it will be easier to organize the development if we merge at this point and develop the correction logic separately.

PR #46, opened 2024-01-26
Changes to port voltage tuning algorithm based on testing

jonnew:


jonnew:
The other things I want to add to this, but I'm unsure how to, is a `Finally()` operator that turns the port voltage off when acquisition stops. My understanding is that all Actions and Functions gathered by `Configure<thing>.cs` nodes are executed by `context.Configure()` at line 23 of `StartAcquisition`:

https://github.com/neurogears/onix-refactor/blob/69c0d3546fb9eae5565a370fb7cfe9b182c06475/OpenEphys.Onix/OpenEphys.Onix/StartAcquisition.cs#L23

I think this sequence would need to have a Finally() appended to it, and `Configure<thing>.cs` nodes should have the ability to register a `FinalizeLink` Action that is executed in the `Finally()` operator appended to the `StartAcqusition`'s observable sequence.

OpenEphys.Onix/OpenEphys.Onix/ConfigureFmcLinkController.cs
```diff
@@ -58,24 +70,29 @@ public override IObservable<ContextTask> Process(IObservable<ContextTask> source
...
-                    device.WriteRegister(FmcLinkController.PORTVOLTAGE, 0);
-                    Thread.Sleep(WaitUntilVoltageSettles);
-                    device.WriteRegister(FmcLinkController.PORTVOLTAGE, voltage);
-                    Thread.Sleep(WaitUntilVoltageSettles);
+
+                    context.WriteRegister(deviceAddress, FmcLinkController.PORTVOLTAGE, 0);
+                    Thread.Sleep(VoltageOffSettleMillis);
+                    context.WriteRegister(deviceAddress, FmcLinkController.PORTVOLTAGE, voltage);
+                    Thread.Sleep(VoltageOnSettleMillis);
+                    context.WriteRegister(deviceAddress, FmcLinkController.PORTVOLTAGE, 33); // Actively pull voltage to lowest level before turning off
```
aacuevas, line 82: Are you sure this line should be here? It's setting the voltage to the minimum before checking the link state

jonnew: lol. yeah thats not right. I will correct

glopesdev:
@jonnew can you add the dispose comments on a separate issue? I want to present and discuss the dispose model in more detail when we meet and would be useful to track these independently.

There are implementation issues with the current logic that I will fix but the architecture is sound. I agree a dispose for link might be in order, will explain a proposal on how we might do it at next meeting.

jonnew:
Hey @glopesdev: I realized that when writing the comment and put them here https://github.com/neurogears/onix-refactor/issues/47. Is that what you are referring to or something different?

glopesdev:
@jonnew yup, just trying to settle on all core functionality and will then merge / rebase this from main, but can try to merge early in that case.

glopesdev:
@jonnew I have split this PR so we can evaluate and merge each of the individual features independently:
- #65 
- #66 
- #67 

If this sounds good, I think we can close this PR without merging for now (keep your local branch just in case, though I also have a backup).

PR #51, opened 2024-02-19
Add harp sync device configuration and data source

### LINKED ISSUE #50: Develop integration for Harp sync input device
There seems to be no official ONI datasheet for the Harp sync input device, but reverse-engineering the ONIX package and using `oni-repl` has revealed the following details on config registers and data frame:

|   |   |
| - | - |
| ID | 30 |
| Address | 12 |
| Enable | 0 |

```c#
struct HarpSyncInputPayload
{
    public ulong HubClock;
    public uint HarpTime;
}
```

glopesdev:
This PR adds configuration and streaming operators for the Harp sync input device on the ONIX board. Default address and device ID were reverse engineered using `oni-repl`. The payload structure was reverse engineered from the previous ONIX plugin.

Fixes #50

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureHarpSyncInput.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureHarpSyncInput.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.ComponentModel;
+
+namespace OpenEphys.Onix
+{
+    public class ConfigureHarpSyncInput : SingleDeviceFactory
+    {
+        public ConfigureHarpSyncInput()
+            : base(typeof(HarpSyncInput))
+        {
+            DeviceAddress = 12;
+        }
+
+        [Category(ConfigurationCategory)]
+        [Description("Specifies whether the Harp sync input device is enabled.")]
+        public bool Enable { get; set; } = true;
+
+        public override IObservable<ContextTask> Process(IObservable<ContextTask> source)
+        {
+            var deviceName = DeviceName;
+            var deviceAddress = DeviceAddress;
+            return source.ConfigureDevice(context =>
+            {
+                var device = context.GetDeviceContext(deviceAddress, HarpSyncInput.ID);
+                device.WriteRegister(HarpSyncInput.ENABLE, Enable ? 1u : 0);
+                return DeviceManager.RegisterDevice(deviceName, device, DeviceType);
+            });
+        }
+    }
+
+    static class HarpSyncInput
+    {
+        public const int ID = 30;
+
+        // managed registers
+        public const uint ENABLE = 0x0; // Enable or disable the data stream
+
+        internal class NameConverter : DeviceNameConverter
+        {
+            public NameConverter()
+                : base(typeof(HarpSyncInput))
+            {
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


glopesdev:
@jonnew @aacuevas let me know if anything is obviously off, we will test it soon anyway. I am aware the hub clock property does not follow the updated convention. This is deliberate since I want to do that renaming in bulk for all devices in a separate PR, but we were in need of testing this feature now.

aacuevas:
The current hardware does not have the capability to switch between controller or receiver. The HarpSyncInput device is always an input.
Old breakout boards have a standalone controller for their HARP output jack.
On newer boards, the jack is an input, and can only act as an input, so the controller was disabled altogether.

What the HarpSyncInput in the latest firmware has, is a register to select the input from either the old, "hacky" way of plugging a dongle into clkin_1, or using the proper new breakout input jack. This is selected by address 1, a value of 0 means the new breakout HARP input jack, a value of 1 means the old dongle on clkin_1

Also, in the newest firmware, the byte order for both the hub clock and the harp word is fixed on this device (and most of the host/hub0 devices)

glopesdev:
> The current hardware does not have the capability to switch between controller or receiver. The HarpSyncInput device is always an input.

Sounds great, thanks for the clarification. We can discuss whether to expose this register as a manual property or do some kind of auto-detection based on the context version.

> Also, in the newest firmware, the byte order for both the hub clock and the harp word is fixed on this device (and most of the host/hub0 devices)

@aacuevas This is really good to know! Does that mean I should be able to just read the harp time directly as a `uint` then? Need to try it on the newer board with the new firmware. Later we can try it in the old board but will need to get a firmware update for hub hardware version 0.4.

aacuevas:
> Sounds great, thanks for the clarification. We can discuss whether to expose this register as a manual property or do some kind of auto-detection based on the context version.
> 
> >

There is no way to know which breakout board is connected. And even with a newer breakout, one could chose (for whatever reason) to use the hacky dongle, so I'd expose it as a property

> Does that mean I should be able to just read the harp time directly as a uint then?
>>

With an up-to-date firmware, yes. And HubClock as a uint64. The same should be true for all host devices. (we fixed this in all devices, but updating firmware in all existing headstages was not feasible. Will be proper for any new devices we make)

> Later we can try it in the old board but will need to get a firmware update for hub hardware version 0.4.
>>

Not sure what do you mean by this

glopesdev:
> Not sure what do you mean by this

Sorry this was my confusion, I thought the change was in the host hardware but realizing now it's the actual breakout board. I thought maybe the "hub hardware revision" could be somehow connected to this, but I guess not.

> There is no way to know which breakout board is connected. And even with a newer breakout, one could chose (for whatever reason) to use the hacky dongle, so I'd expose it as a property

Sounds good

> With an up-to-date firmware, yes. And HubClock as a uint64. The same should be true for all host devices. (we fixed this in all devices, but updating firmware in all existing headstages was not feasible. Will be proper for any new devices we make)

Ok, let's discuss how to propagate this change to the data decoding nodes. Ideally it would be best if we can assume that all clocks are read the same way (the new one). If that's not possible, I'm almost tempted to just make some kind of "legacy" package containing just data streaming nodes for "old" headstages.

aacuevas:
> Ok, let's discuss how to propagate this change to the data decoding nodes. Ideally it would be best if we can assume that all 
> clocks are read the same way (the new one). If that's not possible, I'm almost tempted to just make some kind of "legacy" 
> package containing just data streaming nodes for "old" headstages
>>

My take on this would be to, indeed, assume that all clocks will use the new (proper) format. We could make a set of "legacy" nodes but, IMHO, and given that the HubSyncCounter is seldom used, we could just make an "old HubSyncCounter format correction node" that can be plugged after data collection from an old headstage and call it a day.

PR #56, opened 2024-02-26
Add memory usage monitor support

jonnew:
- Configuration
- High-level data frame and MemoryUsage reporting node

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureBreakoutBoard.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureBreakoutBoard.cs
@@ -11,10 +11,14 @@ namespace OpenEphys.Onix
         [TypeConverter(typeof(HubDeviceConverter))]
         public ConfigureAnalogIO AnalogIO { get; set; } = new();
 
+        [TypeConverter(typeof(HubDeviceConverter))]
+        public ConfigureMemoryMonitor MemoryMonitor { get; set; } = new();
+
         internal override IEnumerable<IDeviceConfiguration> GetDevices()
         {
             yield return Heartbeat;
             yield return AnalogIO;
+            yield return MemoryMonitor; 
         }
     }
 }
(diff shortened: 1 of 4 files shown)


glopesdev:
@jonnew this looks great, I would like to merge this but just to confirm the naming. In the datasheet this device is called [Memory Usage Monitor](https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/memory-usage.html) although the URL is just `memory-usage`.

In your implementation you used two different variations of the term, i.e. `ConfigureMemoryMonitor` (the config node) and `MemoryUsage` (the data node). Do you have any preference on which term would make most sense? For consistency, I believe the choice should be between the following pairs of node names:

### MemoryMonitor
- `ConfigureMemoryMonitor`
- `MemoryMonitorData`

### MemoryUsage
- `ConfigureMemoryUsage`
- `MemoryUsageData`

### MemoryUsageMonitor
- `ConfigureMemoryUsageMonitor`
- `MemoryUsageMonitorData`

jonnew:
Hey cool. I think the first two are fine.

ConfigureMemoryMonitor
MemoryMonitorData

The reason I didnt do this is because AnalogInput didnt have the Data suffix. I guess this will change?

jonnew:
Also, I just realized:

```
        public ulong FrameClock { get; private set; }

        public uint DeviceAddress { get; private set; }
```

Do these need private set since they are assigned in constructor?

glopesdev:
> I think the first two are fine.

Great, will edit directly in the PR branch if you don't mind.

> The reason I didnt do this is because AnalogInput didnt have the Data suffix. I guess this will change?

Yup, let's think about it, it might make sense to change it.

> Do these need private set since they are assigned in constructor?

No need, the syntax with a single `get` means exactly that, you can still assign in the constructor.

glopesdev:
Actually, I will merge this as-is for now, and we can discuss the naming conventions as a whole before release.

glopesdev (approved):

PR #65, opened 2024-03-27
Update FmcLinkController port voltage algorithm

### LINKED ISSUE #64: Port voltage tuning algorithm limitations
Following initial tests on both the 64-channel and RHS2116 headstages, the current port voltage auto-tuning algorithm implemented in `ConfigureFmcLinkController` was found to have significant limitations.

For the 64-channel headstage, the link voltage settle times were found to be on the order of at least 4000 milliseconds. Given the scale of these time intervals, this will significantly impact the time it takes to do port scanning. Since these are electrical properties specific to individual headstage design, they should probably be exposed as additional configurable parameters in the link controller itself so they can be tuned per headstage.

For the RHS2116 the minimum voltage for establishing the SERDES lock (1.8V) is too low to enable full headstage functionality (minimum is 5.0V). This can be partially addressed by the `VoltageOffset` property, but it seems like there are additional limitations on minimum port voltage for resetting the headstage, as well as a potential requirement to do an extra context reset before the headstage can be found, both of which require further investigation.

glopesdev:
- Expose more parameters of its operation
- Final link check after setting offset voltage
- From testing it was found that headstage-64 needs a very large amount of time to reach an acceptable voltage level to be considered off (nearly 10 seconds)
- A new virtual SetPortVoltage method was introduced in the base link controller to allow headstages to change the procedure depending on its needs. In the case of headstage-64, a two-step procedure with a long wait time after turning port voltage off, and a very large offset voltage applied after lock.

Fixes #64

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureFmcLinkController.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureFmcLinkController.cs
@@ -28,13 +28,25 @@ namespace OpenEphys.Onix
         [Description("Specifies the maximum link voltage at which the hub is expected to operate.")]
         public double MaxVoltage { get; set; }
 
+        [Description("Specifies the voltage step size when searching for a SERDES lock.")]
+        public double VoltageIncrement { get; set; } = 0.2;
+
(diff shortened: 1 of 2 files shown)


jonnew:
As I said on discord, this looks like a good solution to me

Issue #58, opened 2024-03-05
Configuration registers with non-null reset action (hs-64 electrical stimulator ENABLE register, specifically)


jonnew:
In [3efaf1b](https://github.com/neurogears/onix-refactor/pull/46/commits/3efaf1be66347fb94c7edc08d1a5ccd4af03111a) I added basic functionality for the electrical stimulation sub-circuit within headstage-64. However,  the trigger node is non-functional because the ENABLE register state is not maintained at 1 after its configured. Namely,

![image](https://github.com/neurogears/onix-refactor/assets/1429475/e02e0810-5948-4019-8913-86c864fb443e)

This is a Configuration property that is set a single time when creating the observable. I have a strong feeling that this is due to ENABLE's reset behavior, which is a bit different than other registers, but is valid under the spec:

![image](https://github.com/neurogears/onix-refactor/assets/1429475/9f05aa29-de25-41d8-8386-36b3bbfe80b6)

(Link to complete register table: https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/estim-hs64.html#register-programming)

The highlighed value is the value that the ENABLE is set to on a reset. The POWERON  register is shown for comparson, where None indicates that the reset has no action on the its value. Most registers behave like POWERON. But for something like enabling the electrical stimulator, its important that the register always start fresh at "off". Maybe this just indicates that ENABLE really needs to be an Acquisition-level, dynamic property that is set to true as part of the workflow's logic at runtime.

glopesdev:
> Maybe this just indicates that ENABLE really needs to be an Acquisition-level, dynamic property that is set to true as part of the workflow's logic at runtime.

@jonnew What we have been doing now in similar cases is to have `Configure` operators which can be used to set both initialization time and runtime configuration parameters (see [modify-pulse-pal-configuration](https://bonsai-rx.org/pulsepal/index.html#modify-pulse-pal-configuration)), and also to allow dynamic setting of individual properties (instead of having everything grouped inside a single configuration or runtime node.

What I like about this is that often for dynamic stimulation scenarios you want to start with a stable known config, and then change it as part of the experimental logic, *but without overwriting your original settings*. By having everything going through only one node with externalizable properties, there is a significant risk of someone saving the workflow after running the protocol and losing important settings.

The other nice thing about this approach is that it is possible to design the architecture to make the code completely reusable, so you wouldn't have to rewrite the logic to allow for these different options.

I can show you an example of this next time we meet.

[...]

PR #66, opened 2024-03-31
Add HS64 electrical stim and trigger nodes

### LINKED ISSUE #27: HS64 Electrical Stimulator configuration and streaming
ONI datasheet for the [HS64 Electrical Stimulator](https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/estim-hs64.html) has details on the config registers.

glopesdev:
This PR adds support for the HS64 electrical stimulator device. Currently the trigger node is non-functional since the `ENABLE` register will be set back to zero on reset. Approaches to resolve this are currently being discussed in #58 and may be incorporated here before merging.

Fixes #27

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64.cs
@@ -31,6 +31,10 @@ namespace OpenEphys.Onix
         [TypeConverter(typeof(HubDeviceConverter))]
         public ConfigureTS4231 TS4231 { get; set; } = new() { Enable = false };
 
+        [Category(ConfigurationCategory)]
+        [TypeConverter(typeof(HubDeviceConverter))]
+        public ConfigureHeadstage64ElectricalStimulator ElectricalStimulator { get; set; } = new() { Enable = false };
+
         public PortName Port
         {
             get { return port; }
@@ -42,6 +46,7 @@ namespace OpenEphys.Onix
                 Rhd2164.DeviceAddress = offset + 0;
                 Bno055.DeviceAddress = offset + 1;
                 TS4231.DeviceAddress = offset + 2;
+                ElectricalStimulator.DeviceAddress = offset + 3;
             }
         }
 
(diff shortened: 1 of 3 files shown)


jonnew:
As we were discussing in discord, and related to my comments on #67, I think we probably need to make ConfigureFmcLinkController pure virtual. Then the entire headstage finding procedure needs to be impemented by the implementing class. For instance, for the time being, I changed SetPortVoltage to the following:

``` csharp
protected virtual bool FindPortVoltage(DeviceContext device)
```

With this, implementers are responsible for their own port voltage tuning algorithm and defining what a successful connection means (return value). For instance, in `ConfigureHeadstage64`, I created the following custom port tuning algorithm which seems to work as well as the last one but is reasonably fast so that it can potentially be used in the field. But of course, this is very specific to this headstage and I think that's going to be a common issue.

``` csharp
protected override bool FindPortVoltage(DeviceContext device)
            {
                // TODO: It takes a huge amount of time to get to 0, almost 10 seconds.
                // The best we can do at the moment is drive port voltage to minimum which
                // is an active process and then settle from there to zero volts.

                var minVoltage = (uint)(MinVoltage * 10);
                var maxVoltage = (uint)(MaxVoltage * 10);
                var voltageOffset = (uint)(VoltageOffset * 10);
                var voltageIncrement = (uint)(VoltageIncrement * 10);

                // Start with highest voltage
                device.WriteRegister(FmcLinkController.PORTVOLTAGE, maxVoltage);
                Thread.Sleep(200);
                if (!CheckLinkState(device)) return false;

                // Ramp it down to find lowest lock voltage
                uint voltage;
                for (voltage = maxVoltage; voltage > minVoltage && CheckLinkState(device); voltage -= voltageIncrement)
                {
                    device.WriteRegister(FmcLinkController.PORTVOLTAGE, voltage);
                    Thread.Sleep(200);
                }

                device.WriteRegister(FmcLinkController.PORTVOLTAGE, 33);
                device.WriteRegister(FmcLinkController.PORTVOLTAGE, 0);
                Thread.Sleep(1000);
                device.WriteRegister(FmcLinkController.PORTVOLTAGE, voltage + voltageOffset);
                Thread.Sleep(200);
                return CheckLinkState(device);

            }
```

jonnew:
This works well and was tested on hardware without issue. Feel free to merge.

PR #67, opened 2024-03-31
Support RHS2116 streaming and stimulation

### LINKED ISSUE #26: RHS2116 configuration and streaming
ONI datasheet for the [RHS2116](https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/rhs2116.html) has details on the config registers and data frame.

glopesdev:
This PR adds basic support for RHS2116 configuration, amplifier data streaming and stimulation functionality.

- Support acquisition from RHS2116, writing filter settings, etc.
- Add stimulus sequence definition to RHS2116 configuration
- Will require GUI support for configuring the stimulus waveform.
- Tested with hardcoded stimulus parameters and it works great.

Fixes #26


OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstageRhs2116.cs
```diff
@@ -0,0 +1,92 @@
...
+            yield return StimulusTrigger;
+        }
+
+        class ConfigureHeadstageRhs2116LinkController : ConfigureFmcLinkController
+        {
+            protected override void SetPortVoltage(DeviceContext device, uint voltage)
+            {
+                // TODO: We check if we have SERDES lock here,
+                // which means we are setting the final voltage offset.
+                var hasLock = CheckLinkState(device);
```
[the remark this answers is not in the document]
jonnew: This is the function that I suggest for the time being. we have a `bool finalVoltage` which defaults to false in base class but can be used to inform  the function that this will be the last voltage value before acquisition. Nothing to do with lock or not. 

``` csharp
 protected override void SetPortVoltage(DeviceContext device, uint voltage, bool finalVoltage)
            {
                // TODO: We check if we have SERDES lock here,
                // which means we are setting the final voltage offset.
                

                const int WaitUntilVoltageOffSettles = 500;
                const int WaitUntilVoltageOnSettles = 500;
                device.WriteRegister(FmcLinkController.PORTVOLTAGE, 0);
                Thread.Sleep(WaitUntilVoltageOffSettles);
                device.WriteRegister(FmcLinkController.PORTVOLTAGE, voltage);
                Thread.Sleep(WaitUntilVoltageOnSettles);

                //var hasLock = CheckLinkState(device);

                // TODO: The RHS2116 headstage needs an additional reset after power on
                // to provide its device table.
                if (finalVoltage)
                {
                    device.Context.Reset();
                    Thread.Sleep(100);
                }
            }
````

File: OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstageRhs2116.cs [hunk 2]
```diff
@@ -0,0 +1,92 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel;
+using System.Threading;
+
+namespace OpenEphys.Onix
+{
+    public class ConfigureHeadstageRhs2116 : HubDeviceFactory
+    {
+        PortName port;
+        readonly ConfigureFmcLinkController LinkController = new();
```
jonnew, line 10: Should be `ConfigureHeadstageRhs2116LinkController`

OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstageRhs2116.cs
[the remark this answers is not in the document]
jonnew: (The 500 500 were needed to make it work reliably)

jonnew (commented):
Overall, I think this is move in the right direction. However, base virtual functions return void and are therefore ripe for redundancy and abuse. It makes me wonder if we should really just virtualize the entire headstage finding procedure and tell the headstage to figure it out and let us know if it finds anything or not.

OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstageRhs2116.cs
```diff
@@ -65,40 +75,38 @@ class ConfigureHeadstageRhs2116LinkController : ConfigureFmcLinkController
...
-                        if (CheckLinkState(device))
-                        {
-                            // TODO: The RHS2116 headstage needs an additional reset after power on
-                            // to provide its device table.
-                            Thread.Sleep(500);
-                            device.Context.Reset();
-                            return true;
-                        }
-                        else break;
+                        return CheckLinkState(device);
```
glopesdev: Just noticed this now, and was really confused. If we just need to reset before the last check, why not simply add `device.Context.Reset()` before the return call?

My suspicion is that perhaps this is related to the case where we want to just set a port voltage override and therefore the code does not go through this path?

In that case, maybe the best would be to mark `ConfigurePortVoltageOverride` itself as `virtual` and override that?

Bonsai/Bonsai.config
```diff
@@ -43,26 +46,31 @@
...
+    <AssemblyLocation assemblyName="MathNet.Numerics" processorArchitecture="MSIL" location="Packages\MathNet.Numerics.5.0.0\lib\net48\MathNet.Numerics.dll" />
     <AssemblyLocation assemblyName="Microsoft.Web.WebView2.Core" processorArchitecture="MSIL" location="Packages\Microsoft.Web.WebView2.1.0.1823.32\lib\net45\Microsoft.Web.WebView2.Core.dll" />
     <AssemblyLocation assemblyName="Microsoft.Web.WebView2.WinForms" processorArchitecture="MSIL" location="Packages\Microsoft.Web.WebView2.1.0.1823.32\lib\net45\Microsoft.Web.WebView2.WinForms.dll" />
     <AssemblyLocation assemblyName="Microsoft.Web.WebView2.Wpf" processorArchitecture="MSIL" location="Packages\Microsoft.Web.WebView2.1.0.1823.32\lib\net45\Microsoft.Web.WebView2.Wpf.dll" />
     <AssemblyLocation assemblyName="OpenCV.Net" processorArchitecture="MSIL" location="Packages\OpenCV.Net.3.4.2\lib\net462\OpenCV.Net.dll" />
+    <AssemblyLocation assemblyName="OpenEphys.PhaseDetector" processorArchitecture="MSIL" location="Packages\OpenEphys.PhaseDetector.0.6.0\lib\net472\OpenEphys.PhaseDetector.dll" />
```
glopesdev, line 73: Is the phase detector package used in the implementation, or just for examples?

jonnew: Oops. Should not be there. Also don't know why I committed this file.

OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstageRhs2116.cs
jonnew: I think this was addressed in #124 which does what you suggest.

OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstageRhs2116.cs
bparks13, line 1: This file (along with a few others in the same OpenEphys.Onix/OpenEphys.Onix folder path) are leftovers from the old folder pattern. These should be removed.

OpenEphys.Onix1/ConfigureRhs2116.cs
```diff
@@ -0,0 +1,265 @@
...
+                    format &= ~0xFu;
+                    format |= (uint)dspCutoff;
+                }
+
+                device.WriteRegister(Rhs2116.FORMAT, format); // NB: DC data only provided in unsigned. Force amplifier data to use unsigned for consistency
+                device.WriteRegister(Rhs2116.ENABLE, enable ? 1u : 0);
+
+                return new CompositeDisposable(
+                    DeviceManager.RegisterDevice(deviceName, device, DeviceType),
+                    analogLowCutoff.Subscribe(newValue =>
```
bparks13, line 108: All calls should be using the SubscribeSafe method instead of Subscribe.

Issue #74, opened 2024-04-11
Optional override of port-voltage discovery algorithm


jonnew:
We took steps to tailor the idiosyncrasies associated with setting the port voltage for different headstages by making ConfigureFmcLinkController.ConfigurePortVoltage abstract:

https://github.com/neurogears/onix-refactor/blob/db83496b195793ab9395d551b3b2882e18caa7f6/OpenEphys.Onix/OpenEphys.Onix/ConfigureFmcLinkController.cs#L24) 

and allowing different headstages to define custom algorithms for tuning the port voltage. However. there are some circumstances where the user really must be given complete control over the voltage. To deal with this, we need to have an optional, nullable property that, when not null, overrides the algorithm and forces a voltage. This can be headstage specific if only so that a headstage-specific description can be provided to tell the user what voltages to aim for.

PR #75, opened 2024-04-11
Add PortVoltage property to ConfigureHeadstage64

jonnew:
- Attempt at #74
- Feels pretty redundant to have to define two properties, one of which just fowards information to and from the other

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64.cs
@@ -50,6 +50,15 @@ namespace OpenEphys.Onix
             }
         }
 
+        [Description("If defined, overrides automated voltage discovery and applies " +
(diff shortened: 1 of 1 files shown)


jonnew:
Once we work this out, I will do similar to the other headstages.

OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64.cs
```diff
@@ -50,6 +50,15 @@ public PortName Port
             }
         }
 
+        [Description("If defined, overrides automated voltage discovery and applies " +
+            "the specified voltage to the headstage. Warning: this device requires 5.5V to 6.0V " +
+            "for proper operation. Higher voltages can damage the headstage.")]
```
glopesdev, line 55: Consult the documentation for allowable voltage ranges. Voltages which are too high may result in damage to devices.

PR #78, opened 2024-04-17
HeadstageRhs2116 - Add GUI element for stimulus sequence, move acquisition parameters

bparks13:
This pull request is to add GUI functionality to the stimulus sequence editor for the **HeadstageRhs2116** device. As part of adding a GUI, parameters that are able to be changed during acquisition are moved from the `ConfigureRhs2116Trigger` node to the `Rhs2116Trigger` node.

One major addition in this PR is the OpenEphys.Onix.Design project, which will handle the GUI element side of things.


jonnew:
- If i double click the HeadstageRjs2116 node, an execption is thrown because the stimulus sequence is empty and zed graph is trying to plot something.

image


image.

bparks13:
Commit [c4ad9ef](https://github.com/neurogears/onix-refactor/pull/78/commits/c4ad9ef3f3eb34b52eb67a6453764aeb0a18c2ac) introduces a class to hold two stimulus sequences together (`Rhs2116StimulusSequenceDual`). However, it does not fully abstract the stimulus sequence, particularly with the Dictionary. If possible, we need to find a way to combine the underlying dictionaries together in a way that does not result in duplicate keys, but allows for the individual device tables to be easily accessed and then written to the appropriate registers.

Might be worth generalizing this function to contain a List or IEnumerable of `Rhs2116StimulusSequence` objects, allowing for N sequences to be handled together.

Issue #14, opened 2023-08-29
Develop integration for NP 2.0 headstage


glopesdev:
- [x] #36
- [x] #37
- [x] #38
- [x] #86
- [x] #87

jonnew:
Re opening so that the commits to issue-14 make more sense

jonnew:
There is a lot of common code between Beta and public variants of the headstage. Specifically most, if not all, of the configuration is the same (shift register, register addresses, gain correction file contents and application, etc). What should we do about this? I feel like `static class NeuropixelsV2eBeta` should only hold things that are different from the public version (e.g. the GPIOConfig defaults) and everything else should go in `static class NeuropixelsV2e`. What do we think about that?

PR #93, opened 2024-05-16
Neuropixels 2.0 configuration

jonnew:
- Bring public version of probe up to speed with beta version
- Add gain calibration files and consequent fixed point multiplication to the data stream
- This commit kinda illustrates the configuration reuse stuff I was talking about in #14 comment section
- This is really more of a code sanity request than a true PR

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.ComponentModel;
 using System.Reactive.Disposables;
+using Bonsai;
 
 namespace OpenEphys.Onix
 {
@@ -14,6 +15,20 @@ namespace OpenEphys.Onix
         [Category(ConfigurationCategory)]
         [Description("Specifies whether the NeuropixelsV2 device is enabled.")]
         public bool Enable { get; set; } = true;
+
+        [Category(ConfigurationCategory)]
+        [Description("Specifies which reference to use for all probe electrodes.")]
+        public NeuropixelsV2eBetaChannelReference Reference { get; set; }
+
+        [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
+        [Description("Path to the Neuropixels 2.0 gain calibraiton file for probe A.")]
+        [Editor("Bonsai.Design.OpenFileNameEditor, Bonsai.Design", DesignTypes.UITypeEditor)]
+        public string GainCalibrationFileA { get; set; }
+
+        [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
+        [Description("Path to the Neuropixels 2.0 gain calibraiton file for probe B.")]
+        [Editor("Bonsai.Design.OpenFileNameEditor, Bonsai.Design", DesignTypes.UITypeEditor)]
+        public string GainCalibrationFileB { get; set; }
 
         public override IObservable<ContextTask> Process(IObservable<ContextTask> source)
         {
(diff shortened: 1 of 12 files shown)


OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs
```diff
@@ -165,6 +206,20 @@ static void ConfigureProbeStreaming(I2CRegisterContext i2cNP)
         }
     }
 
+    class NeuropixesV2eBetaDeviceInfo : DeviceInfo
+    {
+        public NeuropixesV2eBetaDeviceInfo(ContextTask context, Type deviceType, uint deviceAddress, double? gainCorrectionA, double? gainCorrectionB)
+            : base(context, deviceType, deviceAddress)
+        {
+            GainCorrectionA = gainCorrectionA != null ? (ushort)(gainCorrectionA * (1 << 14)) : null;
```
glopesdev, line 214: Is the `1 << 14` conversion necessary here? It looks like the parse already does this and casts it to `ushort` inside the `ReadGainCorrection` method?

Maybe this could be changed to receive a `ushort?` instead?

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs
```diff
@@ -222,6 +277,6 @@ public enum NeuropixelsV2eBetaChannelReference : uint
     {
         External,
         Tip,
-        Internal
+        Ground
```
glopesdev, line 280: We could move shared enums into separate files.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBetaHeadstage.cs
```diff
@@ -35,6 +35,15 @@ public PortName Port
             }
         }
 
+        [Description("If defined, overrides automated voltage discovery and applies " +
+            "the specified voltage to the headstage. Warning: this device requires 3.0V to 5.5V " +
+            "for proper operation. Higher voltages can damage the headstage.")]
+        public double? PortVoltage
```
glopesdev, line 41: I wonder if we should move this to be part of an abstract class instead.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eLinkController.cs
```diff
@@ -4,33 +4,56 @@ namespace OpenEphys.Onix
 {
     class ConfigureNeuropixelsV2eLinkController : ConfigureFmcLinkController
     {
+        public double? PortVoltage { get; set; } = null;
```
glopesdev, line 7: I wonder if we should move this to be part of a common abstract class.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs
jonnew: There is a lot of shared stuff between these two probes. I actually think these should come in the same bonsai package with a lot of shared helpers because they are different versions of the same probe.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBetaHeadstage.cs
jonnew: The only thing that would need an override is the Description. The voltage is headstage dependent.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
```diff
@@ -15,6 +16,20 @@ public ConfigureNeuropixelsV2e()
         [Description("Specifies whether the NeuropixelsV2 device is enabled.")]
         public bool Enable { get; set; } = true;
 
+        [Category(ConfigurationCategory)]
+        [Description("Specifies which reference to use for all probe electrodes.")]
+        public NeuropixelsV2eBetaChannelReference Reference { get; set; }
+
+        [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
+        [Description("Path to the Neuropixels 2.0 gain calibraiton file for probe A.")]
```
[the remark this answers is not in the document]
jonnew: I think explicit is best.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs
jonnew: oops, this is true.

glopesdev:
Superseded by #113

PR #108, opened 2024-06-18
Add DigitalIO device configuration and data nodes

### LINKED ISSUE #80: FMC Host Digital IO configuration and streaming
ONI datasheet for [FMC Digital IO Device](https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/fmc-digital-io.html) has details on data format and control registers.

- Buffering input data, assuming whatever rate coming from the DIO
  - in the future we might consider max rate of 1 MHz at firmware level
  - don't buffer and just return values as fast as possible (we will do this)
- Output `Mat` type vs `bool[]` vs `Enum`
- Consider data type conversions for output sink

glopesdev:
- Adds support for device configuration 
- Introduce `DigitalInputDataFrame` to breakout the device to host payload
- Add DigitalIO device to `ConfigureBreakoutBoard` hub device

Fixes #80

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureBreakoutBoard.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureBreakoutBoard.cs
@@ -12,12 +12,16 @@ namespace OpenEphys.Onix
         public ConfigureAnalogIO AnalogIO { get; set; } = new();
 
         [TypeConverter(typeof(HubDeviceConverter))]
+        public ConfigureDigitalIO DigitalIO { get; set; } = new();
+
+        [TypeConverter(typeof(HubDeviceConverter))]
         public ConfigureMemoryMonitor MemoryMonitor { get; set; } = new();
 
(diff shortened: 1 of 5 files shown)


glopesdev:
@jonnew The `HasFlag` operator can be used to pull the state of individual bits in the enums.

jonnew:
I'm trying to test this but its not working as I would expect. This is my workflow:

image

When I press buttons frames are received, and make to this call:

image

But in the workflow, no frames are show in the visualizer or checked with the `HasFlag` operator.

image

glopesdev:
@jonnew This was my last minute attempt at refactoring gone wrong. This operator can be implemented using a simple projection, but when changing it didn't realize I was still inside `Observable.Create`. It should all work now.

jonnew:
I reviewed this and made some corrections. If you agree with these its good to go. There is logic for grouping the Power and Buttons into one enum: port state actually records the position of the power switches for the ports. So this enum encapsulates all switch/button input from the breakout.

PR #109, opened 2024-06-18
Add reusable helper function to find device nodes

### LINKED ISSUE #102: Provide common functionality for finding device configuration nodes
In order to provide maximum flexibility for interlinking data and control nodes with device configuration nodes, it will be necessary to retrieve configuration node objects from the workflow by their unique identifier. This can mimic the way other resource converters in the language work by recursively navigating the workflow.

glopesdev:
On occasion it may be useful for device data node editors to access properties of the device configuration. Currently there is no explicit link between configuration and data nodes other than at runtime through the device manager.

This PR refactors and exposes the routines used internally by `DeviceNameConverter` to find matching device configuration nodes as helper methods which can in the future be used by more sophisticated data stream configuration editors.

Fixes #102

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/DeviceHelper.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/DeviceHelper.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Bonsai;
+using Bonsai.Expressions;
+
+namespace OpenEphys.Onix
+{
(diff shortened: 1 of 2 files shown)


glopesdev:
@bparks13 this is the long overdue refactoring of the code for finding device configuration nodes.

@jonnew @aacuevas There is one important consideration in that all device configuration dialogs are currently editing directly the configure nodes, so whether this is still necessary remains to be seen. We can leave the refactoring internal though so we can experiment a bit with it in the prototypes.

PR #110, opened 2024-06-18
Group frame distribution by device address

### LINKED ISSUE #62: Refactor FrameReceived to group frame distribution by device address

glopesdev:
Device data nodes need to filter each frame by address before parsing frame contents. Because all data nodes in a workflow are operating in parallel, this means that naively each frame will be checked once by each data node.

This PR implements a more efficient distribution strategy, whereby the context task preemptively groups the frame sequence by device address and provides the sequence of groups to data nodes. With this strategy, each data node only needs to find the group with the right address at subscription time, and can then proceed to process all frames without the need for further tests.

In other words, this moves testing address complexity from `O(n * m)` to `O(n) + O(m)`, where `n` is number of frames, and `m` is number of data nodes. Given that `n` is often in the order of a few thousands per second, this has the potential to result in significant speedups, even for a small number of data nodes.

Fixes #62

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/Bno055Data.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/Bno055Data.cs
@@ -16,8 +16,8 @@ namespace OpenEphys.Onix
             return DeviceManager.GetDevice(DeviceName).SelectMany(deviceInfo =>
             {
                 var device = deviceInfo.GetDeviceContext(typeof(Bno055));
-                return deviceInfo.Context.FrameReceived
-                    .Where(frame => frame.DeviceAddress == device.Address)
+                return deviceInfo.Context
+                    .GetDeviceFrames(device.Address)
                     .Select(frame => new Bno055DataFrame(frame));
             });
         }
--- a/OpenEphys.Onix/OpenEphys.Onix/BreakoutAnalogInput.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/BreakoutAnalogInput.cs
@@ -76,8 +76,8 @@ namespace OpenEphys.Onix
                         },
                         observer.OnError,
                         observer.OnCompleted);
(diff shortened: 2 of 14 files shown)


jonnew:
I tested with the following hardware
- headstage-64
  - Streaming works fine
  - Turn off midstream and recovery is fine
- headstage-np2.0beta
  - Streaming works fine
  - Turn off midstream and recovery is fine

jonnew:
Before merging, let me perform one more test: I notice hardware memory accumulation with BlockRead = 2048 on my computer when using npx 2.0 beta. I want to see if this the case with the old version.

glopesdev:
Sounds good, if you have time you can also compare against JIT optimisation by running in Release mode without debugger (Ctrl+F5).

jonnew:
Interestingly, it seems like it _might_ be a bit worse in terms of keeping up with data collection. Seeting block read to 2048 bytes, on my computer, makes it so that `main` is barely able to keep up with data production but there are periods of ~0 growth in memory usage. These are less frequent on `issue-62` and growth seems more monotonic. 

main:
image

issue-62:
image

However, this is a super noisy test for tons or reasons and really indicates that the system calls, which are occurring at the same rate in both scenarios are probably the issue. In fact, empirically, we have found that sometimes less efficient code is better able to keep up with data because the scheduler keeps in state closer the CPU. .e.g. a 10 kHz feedback loop with small data packets and no processing to speak of has a faster per frame round trip time on my hardware than a 1 kHz loop that does the same thing.

jonnew:
> Sounds good, if you have time you can also compare against JIT optimisation by running in Release mode without debugger (Ctrl+F5).

Oops thats a good point

jonnew:
With optimizations:
- Now issue-62 has a slight advantage. I think the take home above stands regardless of these noisy measurements and we should merge.

main:
image

Issue-62:
image

PR #113, opened 2024-06-21
Neuropixels 2.0 electrode configuration

### LINKED ISSUE #86: NeuropixelsV2 software gain correction based on probe calibration files

### LINKED ISSUE #87: NeuropixelsV2 channel mapping configuration

jonnew:
- Bring public version of probe up to speed with beta version
- Add gain calibration files and consequent fixed point multiplication to the data stream
- Add `NeuropixelsV2QuadShankProbeConfiguration` as a model to store and apply electrode configuration

Developed as a continuation of #93, which can now be closed.
Moved shift register issues to #118

Fixes #86 
Fixes #87

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
@@ -18,7 +18,7 @@ namespace OpenEphys.Onix
 
         [Category(ConfigurationCategory)]
         [Description("Probe A electrode configuration.")]
-        public NeuropixelsV2QuadShankProbe ProbeConfigurationA { get; set; }
+        public NeuropixelsV2QuadShankProbeConfiguration ProbeConfigurationA { get; set; }
 
         [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
         [Description("Path to the gain calibration file for probe A.")]
@@ -27,7 +27,7 @@ namespace OpenEphys.Onix
 
         [Category(ConfigurationCategory)]
         [Description("Probe B electrode configuration.")]
-        public NeuropixelsV2QuadShankProbe ProbeConfigurationB { get; set; }
+        public NeuropixelsV2QuadShankProbeConfiguration ProbeConfigurationB { get; set; }
 
         [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
         [Description("Path to the gain calibration file for probe B.")]
@@ -86,7 +86,7 @@ namespace OpenEphys.Onix
                     ConfigureProbeStreaming(probeControl);
                 }
 
-                var deviceInfo = new NeuropixesV2eDeviceInfo(context, DeviceType, deviceAddress, gainCorrectionA, gainCorrectionB);
+                var deviceInfo = new NeuropixelsV2eDeviceInfo(context, DeviceType, deviceAddress, gainCorrectionA, gainCorrectionB);
                 var disposable = DeviceManager.RegisterDevice(deviceName, deviceInfo);
                 var shutdown = Disposable.Create(() =>
                 {
@@ -152,7 +152,7 @@ namespace OpenEphys.Onix
         {
             if (gainCalibrationFile == null)
             {
-                throw new ArgumentException("Calibraiton file must be specified.");
+                throw new ArgumentException("Calibration file must be specified.");
             }
 
             System.IO.StreamReader gainFile = new(gainCalibrationFile);
@@ -160,7 +160,7 @@ namespace OpenEphys.Onix
 
             if (probeSerialNumber != sn)
             {
-                throw new ArgumentException($"Probe serial number {probeSerialNumber} does not match calibraiton file serial number {sn}.");
+                throw new ArgumentException($"Probe serial number {probeSerialNumber} does not match calibration file serial number {sn}.");
             }
 
             // Q1.14 fixed point conversion
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs
@@ -22,7 +22,7 @@ namespace OpenEphys.Onix
 
         [Category(ConfigurationCategory)]
         [Description("Probe A electrode configuration.")]
-        public NeuropixelsV2QuadShankProbe ProbeConfigurationA { get; set; } = new NeuropixelsV2QuadShankProbe();
+        public NeuropixelsV2QuadShankProbeConfiguration ProbeConfigurationA { get; set; } = new NeuropixelsV2QuadShankProbeConfiguration();
 
         [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
         [Description("Path to the gain calibraiton file for probe A.")]
@@ -31,7 +31,7 @@ namespace OpenEphys.Onix
 
         [Category(ConfigurationCategory)]
         [Description("Probe B electrode configuration.")]
-        public NeuropixelsV2QuadShankProbe ProbeConfigurationB { get; set; } = new NeuropixelsV2QuadShankProbe();
+        public NeuropixelsV2QuadShankProbeConfiguration ProbeConfigurationB { get; set; } = new NeuropixelsV2QuadShankProbeConfiguration();
 
         [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
         [Description("Path to the gain calibraiton file for probe B.")]
@@ -114,7 +114,7 @@ namespace OpenEphys.Onix
                 // Still its good to get them roughly (i.e. within 10 PCLKs) started at the same time.
                 SyncProbes(serializer, gpo10Config);
 
-                var deviceInfo = new NeuropixesV2eDeviceInfo(context, DeviceType, deviceAddress, gainCorrectionA, gainCorrectionB);
+                var deviceInfo = new NeuropixelsV2eDeviceInfo(context, DeviceType, deviceAddress, gainCorrectionA, gainCorrectionB);
                 var disposable = DeviceManager.RegisterDevice(deviceName, deviceInfo);
                 var shutdown = Disposable.Create(() =>
                 {
--- a/OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2RegisterContext.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2RegisterContext.cs
@@ -15,17 +15,17 @@ namespace OpenEphys.Onix
         {
         }
 
-        public void WriteConfiguration(NeuropixelsV2QuadShankProbe probe)
+        public void WriteConfiguration(NeuropixelsV2QuadShankProbeConfiguration probe)
         {
             var baseBits = GenerateBaseBits(probe);
-            WriteShiftRegister(NeuropixelsV2Definitions.SR_CHAIN5, baseBits[0]);
-            WriteShiftRegister(NeuropixelsV2Definitions.SR_CHAIN6, baseBits[1]);
+            WriteShiftRegister(NeuropixelsV2.SR_CHAIN5, baseBits[0]);
+            WriteShiftRegister(NeuropixelsV2.SR_CHAIN6, baseBits[1]);
 
             var shankBits = GenerateShankBits(probe);
-            WriteShiftRegister(NeuropixelsV2Definitions.SR_CHAIN1, shankBits[0]);
-            WriteShiftRegister(NeuropixelsV2Definitions.SR_CHAIN2, shankBits[1]);
-            WriteShiftRegister(NeuropixelsV2Definitions.SR_CHAIN3, shankBits[2]);
-            WriteShiftRegister(NeuropixelsV2Definitions.SR_CHAIN4, shankBits[3]);
+            WriteShiftRegister(NeuropixelsV2.SR_CHAIN1, shankBits[0]);
+            WriteShiftRegister(NeuropixelsV2.SR_CHAIN2, shankBits[1]);
+            WriteShiftRegister(NeuropixelsV2.SR_CHAIN3, shankBits[2]);
+            WriteShiftRegister(NeuropixelsV2.SR_CHAIN4, shankBits[3]);
 
         }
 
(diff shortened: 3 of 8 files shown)


OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
```diff
@@ -35,25 +54,39 @@ public override IObservable<ContextTask> Process(IObservable<ContextTask> source
...
+                if (probeBMetadata.ProbeSerialNumber != null)
                 {
+                    gainCorrectionB = ReadGainCorrection(GainCalibrationFileB, (ulong)probeBMetadata.ProbeSerialNumber);
                     SelectProbe(serializer, NeuropixelsV2e.ProbeBSelected);
+                    probeControl.WriteConfiguration(ProbeConfigurationB);
                     ConfigureProbeStreaming(probeControl);
                 }
 
-                var deviceInfo = new DeviceInfo(context, DeviceType, deviceAddress);
+                var deviceInfo = new NeuropixesV2eDeviceInfo(context, DeviceType, deviceAddress, gainCorrectionA, gainCorrectionB);
```
glopesdev, line 89: Rename class to `NeuropixelsV2eDeviceInfo`

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
```diff
@@ -109,12 +142,31 @@ static uint EnableProbeSupply(I2CRegisterContext serializer)
...
         {
             SelectProbe(serializer, probeSelect);
             return new NeuropixelsV2eMetadata(serializer);
         }
 
+        static ushort ReadGainCorrection(string gainCalibrationFile, ulong probeSerialNumber)
+        {
+            if (gainCalibrationFile == null)
+            {
+                throw new ArgumentException("Calibraiton file must be specified.");
```
glopesdev, line 155:
```suggestion
                throw new ArgumentException("Calibration file must be specified.");
```

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
```diff
@@ -109,12 +142,31 @@ static uint EnableProbeSupply(I2CRegisterContext serializer)
...
+            }
+
+            System.IO.StreamReader gainFile = new(gainCalibrationFile);
+            var sn = ulong.Parse(gainFile.ReadLine());
+
+            if (probeSerialNumber != sn)
+            {
+                throw new ArgumentException($"Probe serial number {probeSerialNumber} does not match calibraiton file serial number {sn}.");
```
glopesdev, line 163:
```suggestion
                throw new ArgumentException($"Probe serial number {probeSerialNumber} does not match calibration file serial number {sn}.");
```

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
```diff
@@ -15,6 +16,24 @@ public ConfigureNeuropixelsV2e()
...
         public bool Enable { get; set; } = true;
 
+        [Category(ConfigurationCategory)]
+        [Description("Probe A electrode configuration.")]
+        public NeuropixelsV2QuadShankProbe ProbeConfigurationA { get; set; }
+
+        [FileNameFilter("Gain calibration files (*_gainCalValues.csv)|*_gainCalValues.csv")]
+        [Description("Path to the gain calibration file for probe A.")]
+        [Editor("Bonsai.Design.OpenFileNameEditor, Bonsai.Design", DesignTypes.UITypeEditor)]
+        public string GainCalibrationFileA { get; set; }
```
glopesdev, line 26: It might be cleaner to fold gain calibration file paths into the `NeuropixelsV2QuadShankProbe` configuration class.

Also probably best to rename that class as `NeuropixelsV2QuadShankProbeConfiguration` for clarity.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
```diff
@@ -109,12 +142,31 @@ static uint EnableProbeSupply(I2CRegisterContext serializer)
...
         }
 
-        NeuropixelsV2eMetadata ReadProbeMetadata(I2CRegisterContext serializer, byte probeSelect)
+        static NeuropixelsV2eMetadata ReadProbeMetadata(I2CRegisterContext serializer, byte probeSelect)
         {
             SelectProbe(serializer, probeSelect);
             return new NeuropixelsV2eMetadata(serializer);
         }
 
+        static ushort ReadGainCorrection(string gainCalibrationFile, ulong probeSerialNumber)
```
glopesdev, line 151: If the above suggestion is accepted (folding gain calibration file into probe configuration), it might also be cleaner to move this function inside the probe configuration class.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs
glopesdev, line 1: All above comments also apply to the Beta probe configuration.

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2Definitions.cs
```diff
@@ -0,0 +1,40 @@
...
+        ProbeB = 1
+    }
+
+    [Flags]
+    enum NeuropixelsV2Status : uint
+    {
+        SR_OK = 1 << 7 // Indicates the SR chain comparison is OK
+    }
+
+    public static class NeuropixelsV2Definitions
```
glopesdev, line 17:
```suggestion
    static class NeuropixelsV2
```

For consistency with the other device static classes, I would prefer to remove the `Definitions` suffix and internalize this class definition.

OpenEphys.Onix/OpenEphys.Onix/NeuropixesV2eDeviceInfo.cs
glopesdev, line 1: Rename to `NeuropixelsV2eDeviceInfo`

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2eMetadata.cs
```diff
@@ -4,23 +4,23 @@ namespace OpenEphys.Onix
...
-                PartNumber = flexI2C.ReadString(OFFSET_FLEXPN, 20);
-                ProbePartNumber = flexI2C.ReadString(OFFSET_PROBEPN, 20);
+                ProbePartNumber = flexI2C.ReadString(OFFSET_PROBE_PN, 20);
+                ProbeSerialNumber = BitConverter.ToUInt64(flexI2C.ReadBytes(OFFSET_PROBE_SN, 8), 0);
+                FlexPartNumber = flexI2C.ReadString(OFFSET_FLEX_PN, 20);
+                var flexVersion = flexI2C.ReadByte(OFFSET_FLEX_VERSION);
+                var flexRevision = flexI2C.ReadByte(OFFSET_FLEX_REVISION);
+                FlexVersion = $"{flexVersion}.{flexRevision}";
             }
             catch (oni.ONIException ex)
```
glopesdev, line 25: Just a comment to address your earlier question, I think it's not a big deal to catch `oni.ONIException` here since it is not really such a general exception (i.e. it is a specific exception class from `clroni`). If there was an alternative way to catch unconnected probes, that might be preferable, but in any case this is an implementation detail and the scope seems fairly restricted here.

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2eBetaMetadata.cs
```diff
@@ -0,0 +1,44 @@
...
+                var flexRevision = flexI2C.ReadByte(OFFSET_FLEX_REVISION);
+                FlexVersion = $"{flexVersion}.{flexRevision}";
+            }
+            catch (oni.ONIException ex)
+            {
+                const int FailureToReadRegister = -5;
+                if (ex.Number != FailureToReadRegister)
+                {
+                    throw;
+                }
```
glopesdev, line 31: Is there an enum somewhere in `clroni` that documents these error codes? If not, we could consider adding it as an internal enum inside this library temporarily.

This would allow us to directly cast the number into the enum value and match against that instead of adding local constants, e.g. looks like this check could be applied verbatim to the non-beta probe.

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2RegisterContext.cs
```diff
@@ -0,0 +1,143 @@
...
+            for (int i = 0; i < bytes.Length; i++)
+            {
+                // NB: http://graphics.stanford.edu/~seander/bithacks.html
+                bytes[i] = (byte)((bytes[i] * 0x0202020202ul & 0x010884422010ul) % 1023);
+            }
+
+            return bytes;
+        }
+
+        // TODO: NeuropixelsV2Definitions.STATUS always fails.
```
glopesdev, line 53:
```suggestion
        // TODO: NeuropixelsV2.STATUS always fails.
```

If the above recommendation to rename the internal static class is accepted, this should be renamed too.

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2RegisterContext.cs
```diff
@@ -0,0 +1,143 @@
...
+        private void WriteShiftRegister(uint srAddress, BitArray data)
+        {
+            var bytes = BitArrayToBytes(data);
+
+            //var count = 2;
+            //while (count-- > 0)
+            //{
+                // This allows Base shift registers to get a good STATUS, but does not help shank registers.
+                //WriteByte(NeuropixelsV2Definitions.SOFT_RESET, 0xFF);
+                //WriteByte(NeuropixelsV2Definitions.SOFT_RESET, 0x00);
```
glopesdev, line 63:
```suggestion
                //WriteByte(NeuropixelsV2.SOFT_RESET, 0xFF);
                //WriteByte(NeuropixelsV2.SOFT_RESET, 0x00);
```

Same as above

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2RegisterContext.cs
```diff
@@ -0,0 +1,143 @@
...
+                WriteByte(NeuropixelsV2Definitions.SR_LENGTH1, (uint)bytes.Length % 0x100);
+                WriteByte(NeuropixelsV2Definitions.SR_LENGTH2, (uint)bytes.Length / 0x100);
+
+                foreach (var b in bytes)
+                {
+                    WriteByte(srAddress, b);
+                }
+            //}
+
+            //if (ReadByte(NeuropixelsV2Definitions.STATUS) != (uint)NeuropixelsV2Status.SR_OK)
```
glopesdev, line 74:
```suggestion
            //if (ReadByte(NeuropixelsV2.STATUS) != (uint)NeuropixelsV2Status.SR_OK)
```

Same as above

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2QuadShankProbe.cs
glopesdev, line 1: Rename to `NeuropixelsV2QuadShankProbeConfiguration`

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2QuadShankProbe.cs
```diff
@@ -0,0 +1,140 @@
...
+        [XmlIgnore]
+        public int Channel { get; private set; } = 0;
+        [XmlIgnore]
+        public int Shank { get; private set; } = 0;
+        [XmlIgnore]
+        public int IntraShankElectrodeIndex { get; private set; } = 0;
+        [XmlIgnore]
+        public NeuropixelsV2QuadShankBank Bank { get; private set; } = NeuropixelsV2QuadShankBank.A;
+        [XmlIgnore]
+        public PointF Position { get; private set; } = new(0f, 0f);
```
glopesdev, line 138: Are all these auxiliary fields in preparation for the GUI? They don't seem to be functional elsewhere.

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2QuadShankProbe.cs
```diff
@@ -0,0 +1,140 @@
...
+        A,
+        B,
+        C,
+        D,
+    }
+
+    public class NeuropixelsV2QuadShankProbe
+    {
+        [XmlIgnore]
+        public readonly List<NeuropixelsV2QuadShankElectrode> Electrodes;
```
glopesdev, line 27: If this is a true read-only list you could consider exposing it as a `public IReadOnlyList<NeuropixelsV2QuadShankElectrode>` property instead.

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2QuadShankProbe.cs
```diff
@@ -0,0 +1,140 @@
...
+
+    public class NeuropixelsV2QuadShankElectrode
+    {
+        private int electrodeNumber = 0;
+        public int ElectrodeNumber
+        {
+            get
+            {
+                return electrodeNumber;
+            }
```
glopesdev, line 65:
```suggestion
            get => electrodeNumber;
```

Taking advantage of new getter syntax for conciseness.

glopesdev (changes requested):

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
glopesdev: We name the class but decided to leave the calibration file properties to be explicit and consistent with NeuropixelsV1.

OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2e.cs
glopesdev: We rejected it

OpenEphys.Onix/OpenEphys.Onix/NeuropixelsV2eBetaMetadata.cs
glopesdev: Since they are needed so sporadically (and best avoided) we decided to leave them local.

glopesdev (approved):

Issue #112, opened 2024-06-20
Miniscope V4 support


jonnew:
Add support for miniscope 4.0

jonnew:
Things to verify:

1. Intentionally lower voltage and look for connectivity issues or phase shifts in frame data
2. Check the wall time between frames to make sure that FPS matches the setting

PR #115, opened 2024-06-24
Add support for UCLA Miniscope V4

jonnew:
- Addresses #112 
- This device needs some work arounds because its clock is completely out of spec for the SERDES it uses. The back channel must be used to configure a PLL on the camera in order to bump the PCLK rate before it will like. For this reason, I made `ConfigureFmcLinkController.CheckLinkState` virtual so I could override its behavior with these steps before checking the link state
- Added configuration and streaming functionality.
- There are few outstanding issues:
1. Something is wrong with the DS90UB9x configuration resulting in 1 missing column from each frame and imaging data that "slides" by one column for each image produced
2. The axis map for the BNO needs to be set correctly
3. I was unable to verify that the EWL was functioning. It might be fine, but I was unable to verify.

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4.cs
@@ -1,0 +1,54 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel;
+
+namespace OpenEphys.Onix
+{
+    public class ConfigureUclaMiniscopeV4 : HubDeviceFactory
+    {
+        PortName port;
+        readonly ConfigureUclaMiniscopeV4LinkController LinkController = new();
+
+        public ConfigureUclaMiniscopeV4()
+        {
+            Port = PortName.PortA;
+            LinkController.HubConfiguration = HubConfiguration.Passthrough;
+        }
+
+        [Category(ConfigurationCategory)]
+        [TypeConverter(typeof(HubDeviceConverter))]
+        public ConfigureUclaMiniscopeV4Camera Imager { get; set; } = new();
+
+        [Category(ConfigurationCategory)]
+        [TypeConverter(typeof(HubDeviceConverter))]
+        public ConfigureUclaMiniscopeV4Bno055 Bno055 { get; set; } = new();
+
+        public PortName Port
+        {
+            get { return port; }
+            set
+            {
+                port = value;
+                var offset = (uint)port << 8;
+                LinkController.DeviceAddress = (uint)port;
+                Imager.DeviceAddress = offset + 0;
+                Bno055.DeviceAddress = offset + 1;
+            }
+        }
+
+        [Description("If defined, it will override automated voltage discovery and apply the specified voltage" +
+                     "to the headstage. Warning: this device requires 5.0V to 6.0V for proper operation." +
+                     "Supplying higher voltages may result in damage to the headstage.")]
+        public double? PortVoltage
+        {
+            get => LinkController.PortVoltage;
+            set => LinkController.PortVoltage = value;
+        }
+
+        internal override IEnumerable<IDeviceConfiguration> GetDevices()
+        {
+            yield return LinkController;
+            yield return Imager;
+            yield return Bno055;
+        }
+    }
+}
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4.cs
@@ -16,7 +16,7 @@ namespace OpenEphys.Onix
 
         [Category(ConfigurationCategory)]
         [TypeConverter(typeof(HubDeviceConverter))]
-        public ConfigureUclaMiniscopeV4Camera Imager { get; set; } = new();
+        public ConfigureUclaMiniscopeV4Camera Camera { get; set; } = new();
 
         [Category(ConfigurationCategory)]
         [TypeConverter(typeof(HubDeviceConverter))]
@@ -30,7 +30,7 @@ namespace OpenEphys.Onix
                 port = value;
                 var offset = (uint)port << 8;
                 LinkController.DeviceAddress = (uint)port;
-                Imager.DeviceAddress = offset + 0;
+                Camera.DeviceAddress = offset + 0;
                 Bno055.DeviceAddress = offset + 1;
             }
         }
@@ -47,7 +47,7 @@ namespace OpenEphys.Onix
         internal override IEnumerable<IDeviceConfiguration> GetDevices()
         {
             yield return LinkController;
-            yield return Imager;
+            yield return Camera;
             yield return Bno055;
         }
     }
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4Camera.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4Camera.cs
@@ -23,8 +23,16 @@ namespace OpenEphys.Onix
         public bool Enable { get; set; } = true;
 
         [Category(ConfigurationCategory)]
-        [Description("Only turn on excitation LED during camera exposures.")]
+        [Description("Camera video rate in frames per second.")]
         public UclaMiniscopeV4FramesPerSecond FrameRate { get; set; } = UclaMiniscopeV4FramesPerSecond.Fps30Hz;
+
+        [Description("Camera sensor analog gain.")]
+        [Editor(DesignTypes.SliderEditor, typeof(UITypeEditor))]
+        public UclaMiniscopeV4SensorGain SensorGain
+        {
+            get => sensorGain.Value;
+            set => sensorGain.OnNext(value);
+        }
 
         [Category(ConfigurationCategory)]
         [Description("Only turn on excitation LED during camera exposures.")]
@@ -38,14 +46,6 @@ namespace OpenEphys.Onix
         {
             get => ledBrightness.Value;
             set => ledBrightness.OnNext(value);
-        }
-
-        [Description("Camera sensor analog gain.")]
-        [Editor(DesignTypes.SliderEditor, typeof(UITypeEditor))]
-        public UclaMiniscopeV4SensorGain SensorGain
-        {
-            get => sensorGain.Value;
-            set => sensorGain.OnNext(value);
         }
 
         [Description("Liquid lens voltage (Volts RMS).")]
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4Camera.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4Camera.cs
@@ -1,0 +1,197 @@
+﻿using System;
+using System.ComponentModel;
+using System.Reactive.Disposables;
+
+namespace OpenEphys.Onix
+{
+    public class ConfigureUclaMiniscopeV4Camera : SingleDeviceFactory
+    {
+        public ConfigureUclaMiniscopeV4Camera()
+            : base(typeof(UclaMiniscopeV4))
+        {
+        }
+
+        [Category(ConfigurationCategory)]
+        [Description("Specifies whether the camera is enabled.")]
+        public bool Enable { get; set; } = true;
+
+        [Category(ConfigurationCategory)]
+        [Description("Only turn on excitation LED during camera exposures.")]
+        public bool InterleaveLed { get; set; } = false;
+
+        [Category(ConfigurationCategory)]
+        [Description("Only turn on excitation LED during camera exposures.")]
+        public UclaMiniscopeV4FramesPerSecond FrameRate { get; set; } = UclaMiniscopeV4FramesPerSecond.Fps30Hz;
+
+        public override IObservable<ContextTask> Process(IObservable<ContextTask> source)
+        {
+            var enable = Enable;
+            var deviceName = DeviceName;
+            var deviceAddress = DeviceAddress;
+            return source.ConfigureDevice(context =>
+            {
+                // configure device via the DS90UB9x deserializer device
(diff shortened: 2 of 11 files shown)


jonnew:
@aacuevas Please have a look at the ConfigureUclaMiniscopeV4Camera.ConfigureMiniscope function to see if there is anything that would result in the "sliding" image (we seem to be missing 1 row of raw data per image)

OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4Camera.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        internal static void ConfigureMiniscope(DeviceContext device)
+        {
+            // configure deserializer
+            device.WriteRegister(DS90UB9x.TRIGGEROFF, 0);
+            device.WriteRegister(DS90UB9x.READSZ, UclaMiniscopeV4.SensorColumns);
+            device.WriteRegister(DS90UB9x.TRIGGER, (uint)DS90UB9xTriggerMode.HsyncEdgePositive);
+            device.WriteRegister(DS90UB9x.SYNCBITS, 0);
+            device.WriteRegister(DS90UB9x.DATAGATE, (uint)DS90UB9xDataGate.VsyncPositive);
+            device.WriteRegister(DS90UB9x.MARK,0);
```
aacuevas, line 82: This should be DS90UB9xMarkMode.VsyncRising instead of 0

Since each ONI frame is one picture row, we need to know which row is the first on every picture frame. This marks the first row with a specific bit.

OpenEphys.Onix/OpenEphys.Onix/UclaMiniscopeV4ImageData.cs
```diff
@@ -0,0 +1,98 @@
...
+                    var passthrough = device.GetPassthroughDeviceContext(DS90UB9x.ID);
+                    var scopeData = device.Context.FrameReceived.Where(frame => frame.DeviceAddress == passthrough.Address);
+                    return Observable.Create<UclaMiniscopeV4DataFrame>(observer =>
+                    {
+                        var sampleIndex = 0;
+                        var imageBuffer = new short[UclaMiniscopeV4.SensorRows * UclaMiniscopeV4.SensorColumns];
+                        var hubClockBuffer = new ulong[UclaMiniscopeV4.SensorRows];
+                        var clockBuffer = new ulong[UclaMiniscopeV4.SensorRows];
+
+                        var frameObserver = Observer.Create<oni.Frame>(
```
aacuevas, line 69: I am not sure where in this lib it should go, but need a filter to wait until the first row is received. On the old lib we did it with `SkipWhile(f => (f.Sample[5] & 0x8000) == 0)`

Additionally, this "first row" bit needs to be ignored, if it were to cause an issue later (Miniscope data is only 12 bits after all)

aacuevas (commented):
The issue is not that it is skipping rows, but we are not telling the software which row is the first. Since startup is asynchronous and each ONI frame is one picture row, we need to mark the first row of every picture frame so if acquisition starts mid picture (or if there were some transmission error), the software can ignore the first rows until a new picture frame starts, so it can create the image proper.

jonnew:
Thanks for the feedback. While I did see this in the old library, I was a bit confused why it was needed. My understanding is that `device.WriteRegister(DS90UB9x.DATAGATE, (uint)DS90UB9xDataGate.VsyncPositive);` means that, although we might start in the middle of an image, this phase shift should be preserved for all time. Obviously that is still an issue, but I dont get why we seem to be transmitting one too few rows per frame.

I will try these suggestions regardless because I might be misunderstanding.

jonnew:
Here is a GIF illustrating the issue (note that this does not use Pixel Mark). Adding pixel mark would _appear_ to solve this issue by resetting the phase back to 0 for each image, but it looks like data is being lost and this is independent of that.

ezgif-5-5ac264e2dd

OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4Camera.cs
```diff
@@ -0,0 +1,197 @@
...
+            device.WriteRegister(DS90UB9x.SYNCBITS, 0);
+            device.WriteRegister(DS90UB9x.DATAGATE, (uint)DS90UB9xDataGate.VsyncPositive);
+            //device.WriteRegister(DS90UB9x.MARK, (uint)DS90UB9xMarkMode.VsyncRising); // TODO: Not sure why this is required given that data is gated by VSYNC HIGH
+
+            // configure deserializer I2C aliases
+            var deserializer = new I2CRegisterContext(device, DS90UB9x.DES_ADDR);
+            uint coaxMode = 0x4 + (uint)DS90UB9xMode.Raw12BitLowFrequency; // 0x4 maintains coax mode
+            deserializer.WriteByte((uint)DS90UB9xDeserializerI2CRegister.PortMode, coaxMode);
+
+            uint alias = UclaMiniscopeV4.AtMegaAddress << 1;
```
glopesdev, line 89: `alias` is a contextual keyword in C# so I would avoid using it if possible, perhaps `aliasValue`?

OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4LinkController.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            try
+            {
+                var ds90ub9x = device.Context.GetPassthroughDeviceContext(DeviceAddress << 8, DS90UB9x.ID);
+                ConfigureUclaMiniscopeV4Camera.ConfigureMiniscope(ds90ub9x);
+            }
+            catch (oni.ONIException ex)
+            {
+                // this can occur if power is too low, so we need to be able to try again
+                const int FailureToWriteRegister = -6;
```
glopesdev, line 46: Same constant as in #113 regarding having an internal enum to map these error codes, instead of having local constant definitions.

aacuevas:
> Thanks for the feedback. While I did see this in the old library, I was a bit confused why it was needed. My understanding is that `device.WriteRegister(DS90UB9x.DATAGATE, (uint)DS90UB9xDataGate.VsyncPositive);` means that, although we might start in the middle of an image, this phase shift should be preserved for all time. Obviously that is still an issue, but I dont get why we seem to be transmitting one too few rows per frame.
> 
> I will try these suggestions regardless because I might be misunderstanding.

The gate is just that, a gate. It just tells the raw device "ignore anything that happens when vsync is not high", but vsync is high for the entire duration of a pixel frame.
This would not be an issue if an ONI frame were a picture frame, but due to FIFO size constraints, an ONI frame is a picture row. And the important part: the raw device can start sending rows before the Bonsai node subscription happens. If that is the case, then the first rows will be discarded and Bonsai will start to listen mid-picture frame. We need to tell Bonsai to ignore everything until a new picture frame happens, that is what the mark does, set a mark on the oni frames that correspond with the 1st row, so bonsai can wait for it and then reconstruct a full frame.

The ´SkipWhile´ statement in the observer is required so Bonsai skips anything until it knows it's going to have a full picture frame.

OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public ConfigureUclaMiniscopeV4()
+        {
+            Port = PortName.PortA;
+            LinkController.HubConfiguration = HubConfiguration.Passthrough;
+        }
+
+        [Category(ConfigurationCategory)]
+        [TypeConverter(typeof(HubDeviceConverter))]
+        public ConfigureUclaMiniscopeV4Camera Imager { get; set; } = new();
```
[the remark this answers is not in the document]
jonnew: Shoudl be camera. Latest commit started to replace Image instances with Camera, but this one was missing.

OpenEphys.Onix/OpenEphys.Onix/ConfigureUclaMiniscopeV4Camera.cs
jonnew: Oops. used i2cAlias.

jonnew:
FWIW, I used another miniscope PCB and the sliding phase shift in the image data disappeared. This indicates that although we are not guaranteed to start at the correct row, the data integrity is another issue entirely. Some scope PCBs seem to have significant data loss. Not sure if this has to do with miniscope design or the generic DS90UB9x receiver, or some combination of the two. In the meantime I have added the MarkMode so that images are aligned to row zero. The data loss issue is something else though.

aacuevas:
Can we compare the same faulty devices with the old lib? We had never noticed this specific behavior, so it could be that the old lib handled data losses better somehow, that there is still some software issue going on or that we were just lucky then and unlucky now.

OpenEphys.Onix/OpenEphys.Onix/UclaMiniscopeV4CameraData.cs
```diff
@@ -0,0 +1,106 @@
...
+using Bonsai;
+using OpenCV.Net;
+
+namespace OpenEphys.Onix
+{
+    public class UclaMiniscopeV4CameraData : Source<UclaMiniscopeV4DataFrame>
+    {
+        readonly BehaviorSubject<double> ledBrightness = new(0);
+        readonly BehaviorSubject<UclaMiniscopeV4SensorGain> sensorGain = new(UclaMiniscopeV4SensorGain.Low);
+        readonly BehaviorSubject<double> liquidLensVoltage = new(47); // NB: middle of range
```
glopesdev, line 19: Should these be in the data node or the configure node? On one hand I can see the appeal of dynamic configuration while the device is streaming, on the other hand I would expect the configure node to be able to fully configure the sensor even if no downstream data node is immediately used. It also makes it more confusing that settings are distributed across two nodes.

Maybe this is one of the cases where we want to allow for the configure node to write to the device while the workflow is running, or alternatively we could have an editor that previews the camera for configuration.

jonnew: I think I agree. I put them here because I for some reason thought they should belong to the Data node. I think its better not to split configuration.

jonnew:
Aaron suggests testing the same hardware with the deprecated Bonsai.ONIX library to see if there is a different behavior.

OpenEphys.Onix/OpenEphys.Onix/UclaMiniscopeV4Image.cs
glopesdev, line 1: @jonnew What is the motivation to rename `UclaMiniscopeV4DataFrame` to `UclaMiniscopeV4Image`? It wasn't mentioned in the commit message, but it seems confusing since it might give people the wrong impression that you can directly manipulate this type as if it was an image, which is not true.

In fact, the type has an `Image` property which is what effectively gives you the image. It also creates inconsistency with both the rest of the library streaming data, and with other camera packages, e.g. Spinnaker, Pylon, etc which also use the `DataFrame` terminology.

jonnew: The motivation is that cameras produce images. All data types contain data. DataFrame could be BNO data.

glopesdev: The output of the `AnalogInput` node is still [`AnalogInputDataFrame`](https://github.com/neurogears/onix-refactor/blob/main/OpenEphys.Onix/OpenEphys.Onix/AnalogInputDataFrame.cs). In that case the discussion was what to name the "data" node, not the "data frame" type.

The BNO data for the UCLA miniscope is just a `Bno055DataFrame` since that is what comes out of the BNO055 device.

You are right that the Miniscope headstage has multiple devices inside it, but we are naming `DataFrame` types after device names, not after headstage names, and I think there is a hard constraint in ONIX that for a single device there can be at most a single data frame stream, is that right? So I don't think we really have any inconsistency here so far.

jonnew: My point was that in other circumstances, we didn't follow conventions in the case where clarity was at stake. I think UclaMiniscopeV4CameraData is a good compromise.

Remember that the Miniscope is not an ONI device. The ONI device is the link controller in passthrough mode. If we were going to follow this convention rigidly, then Miniscope would not even appear in the type.

In any case,  UclaMiniscopeV4CameraData  is a good compromise, so lets use that.

glopesdev: > Remember that the Miniscope is not an ONI device.

It is not an ONI device, but it is registered in the package DeviceManager as a virtual device. The only reason this is not a real ONI device is an implementation detail due to the constraints of the passthrough mode. Even conceptually from a user-centric perspective I feel it is probably better to call it a device.

I don't think we should use the technical details of passthrough mode devices as an argument to guide our naming conventions. If we did, then we would have to change the other passthrough `DataFrame` types like the NPX 1.0e, 2.0e, etc.

The reason we didn't use the `Data` suffix for these types to begin with is that we use `Data` as a suffix for the data node, as that is more readable in the workflow and we can't have name clashes, so the `NeuropixelsV1eData` node produces a sequence of `NeuropixelsV1eDataFrame` objects.

From looking further in the commit, I see the reason you didn't fall into this name clash is that you named the data node as `UclaMiniscopeV4Camera` instead of `UclaMiniscopeV4CameraData` as per the other devices.

The reason we use the `Data` suffix for data nodes is that we use the device name for the static class containing the device registers. In this case you called this class `UclaMiniscopeV4` which is why you didn't get a name clash there either, but this class really feels like it should be called `UclaMiniscopeV4Camera` since the registers here should be the passthrough registers for the pass-through camera device.

Given the above, I don't think we should agree on this compromise, unless we change our naming convention for all the other devices in the package.

jonnew:
@aacuevas brought a very obvious test to do: Make sure the miniscopes I'm using function correctly in Bonsai.ONIX because its using the same mechanism to collect data. Then look for the deltas between passthrough settings and how data is processed in that library versus this one.

jonnew:
This work was completed in #275

PR #125, opened 2024-07-02
Read device ID from type metadata

### LINKED ISSUE #116: Improve exception messages
This message is not helpful

https://github.com/neurogears/onix-refactor/blob/481a9150953d1893c916aebb1b0dda721362d527/OpenEphys.Onix/OpenEphys.Onix/ContextHelper.cs#L17

because it results in this inscrutable text presented to the end user:

"The selected device is not a 33 device." 

For example. What was "selected"? What is a 33 device?

glopesdev:
The current implementation of `GetDeviceContext` and `GetPassthroughDeviceContext` explicitly receives the device ID as a numerical argument. While functionally sound, this approach makes it harder to generate informative error messages, since there is no way to map the ID back to a device name.

This PR refactors this approach to instead receive the expected device type. The numerical device ID is then extracted directly from the type metadata by searching for a literal field with the name `ID`. If no such field is found, or a device with an invalid ID is detected, the exception will now contain the type name.

The change also establishes parity between `ContextHelper.GetDeviceContext` and `DeviceInfo.GetDeviceContext` as they now both receive types as arguments.

The main caveat of this approach is that now declaring a static class with a `const ID` field is no longer optional. Since the approach relies on type metadata, the compiler cannot enforce the compliance of device types with this rule at compile-time. This is probably minor as any errors or inconsistencies will immediately throw errors on any test. The most significant impact will be refactoring the name of this field in the long-term, which would have to be done with care.

Fixes #116

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureAnalogIO.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureAnalogIO.cs
@@ -119,7 +119,7 @@ namespace OpenEphys.Onix
             var deviceAddress = DeviceAddress;
             return source.ConfigureDevice(context =>
             {
-                var device = context.GetDeviceContext(deviceAddress, AnalogIO.ID);
+                var device = context.GetDeviceContext(deviceAddress, DeviceType);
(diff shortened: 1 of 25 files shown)


jonnew (approved):
Looks good, much more helpful.

PR #126, opened 2024-07-02
Ensure any BNO polling exceptions are routed

### LINKED ISSUE #121: Bonsai closes when losing lock of NP1.0e during acquisition when BNO is isntantiated
While acquisition is active including a np1.0e headstage, if the headstage loses lock, Bonsai crashes and closes with no message.
This has been reproduced on a demo machine running the library outside Visual Studio, so there might be an uncaught exception that maybe running through VS gets caught.

This only happens when the BNO node is instantiated and does not happen when it is not present or with hs64, so it is probably related with the i2c functions used to poll the BNO.

### LINKED ISSUE #128: Exceptions thrown during dynamic property assignment are not routed
The current pattern for dynamic properties which may change at runtime is to use a `BehaviorSubject` routing new value notifications to the hardware. However, because these changes may come asynchronously from any thread, any exceptions raised during handling of those notifications will be raised in the calling thread, and not routed correctly to the source observer.

A possible solution is to implement a variant of `Subscribe` as an extension method which receives the original observer as argument and does provide the correct routing.

glopesdev:
BNO polling was guarded against asynchronous context tear-down, but not against random register access exceptions such as may happen following arbitrary disconnects.

This PR attempts to guard against this by routing any exceptions thrown by the polling action to the downstream observer error channel. This might not have been necessary in other devices since frame distribution may be managed differently by the context task and the lower-level firmware. If this approach is successful, we could easily apply it to all cases.

We also introduce a new overload for `ConfigureDevice` which allows similar protection against any exceptions raised by dynamic configuration properties, as described in #128.

Fixes #121 
Fixes #128

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeartbeat.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeartbeat.cs
@@ -33,11 +33,11 @@ namespace OpenEphys.Onix
         {
             var deviceName = DeviceName;
             var deviceAddress = DeviceAddress;
-            return source.ConfigureDevice(context =>
+            return source.ConfigureDevice((context, observer) =>
             {
                 var device = context.GetDeviceContext(deviceAddress, Heartbeat.ID);
                 device.WriteRegister(Heartbeat.ENABLE, 1);
-                var subscription = beatsPerSecond.Subscribe(newValue =>
+                var subscription = beatsPerSecond.SubscribeSafe(observer, newValue =>
                 {
                     var clkHz = device.ReadRegister(Heartbeat.CLK_HZ);
                     device.WriteRegister(Heartbeat.CLK_DIV, clkHz / newValue);
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureLoadTester.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureLoadTester.cs
@@ -40,7 +40,7 @@ namespace OpenEphys.Onix
             var deviceAddress = DeviceAddress;
             var receivedWords = ReceivedWords;
             var transmittedWords = TransmittedWords;
-            return source.ConfigureDevice(context =>
+            return source.ConfigureDevice((context, observer) =>
             {
                 var device = context.GetDeviceContext(deviceAddress, LoadTester.ID);
                 device.WriteRegister(LoadTester.ENABLE, 1);
@@ -59,7 +59,7 @@ namespace OpenEphys.Onix
 
                 var writeArray = Enumerable.Repeat((uint)42, (int)(transmittedWords + 2)).ToArray();
                 device.WriteRegister(LoadTester.HTOD32_WORDS, transmittedWords);
-                var frameHzSubscription = frameHz.Subscribe(newValue =>
+                var frameHzSubscription = frameHz.SubscribeSafe(observer, newValue =>
                 {
                     device.WriteRegister(LoadTester.CLK_DIV, clk_hz / newValue);
                     var max_size = ValidSize();
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV1e.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV1e.cs
@@ -87,15 +87,14 @@ namespace OpenEphys.Onix
                 }
 
                 var deviceInfo = new NeuropixelsV1eDeviceInfo(context, DeviceType, deviceAddress, probeControl);
-                var disposable = DeviceManager.RegisterDevice(deviceName, deviceInfo);
                 var shutdown = Disposable.Create(() =>
                 {
                     serializer.WriteByte((uint)DS90UB9xSerializerI2CRegister.GPIO10, NeuropixelsV1e.DefaultGPO10Config);
                     serializer.WriteByte((uint)DS90UB9xSerializerI2CRegister.GPIO32, NeuropixelsV1e.DefaultGPO32Config);
                 });
                 return new CompositeDisposable(
-                    shutdown,
-                    disposable);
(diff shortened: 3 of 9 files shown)


jonnew:
With both issue-131 merged, turnong off the neuropixels headstages midstream results in hard crash with no exception pop up. With issue-121 on top of this, we get the same behavior but failure to read register exception is presented before the crash. I managed to get a screen grab before everything closes:

image

glopesdev:
@jonnew Can you retry this test with the latest commit?

jonnew:
The latest commit seems to prevent the hard crash. Additionally, the power is turned off to the port so if I try to switch the power back  on, there is no voltage. I think this is a good thing.

However, when i try to restart the workflow it becomes apparaent that maybe things were not fully disposed. 

image

This requires restarted Bonsai to get out of.

glopesdev:
@jonnew Thanks Jon this is really helpful, good to know we're getting there, the last bit should hopefully be easy to fix.

glopesdev:
@jonnew I think I tracked down the source of the issue. The `readFrames` task does not currently route its own exceptions anywhere, i.e. if `ReadFrame` crashes, the task will abort but there will be no-one will be notified. This was true for the old library as well, so not really sure how this was handled before.

I am going to treat this as an irrecoverable error, i.e. if `ReadFrame` crashes then the context should terminate cleanly and do a tear-down. Let me know if you have any other thoughts on this.

glopesdev:
Last issue fixed via #132. @jonnew can you merge issue-131 into your local test branch and see if these two in combination solve the problem? Seems to work fine now on my end.

jonnew:
Unfortunately, I see the same behavior:

image

jonnew:
Unfortunately, I see the same behavior with issue-131 merged:

image

glopesdev:
@jonnew this is unfortunate. Can you attach a stack trace of the original error (i.e. the one that crashes the workflow the first time)?

You can get it either from visual studio, or if you are not easily able to attach in the right place, as long as the editor highlights the correct node and colors the status bar red, you can right-click the status bar at the bottom where the error message is displayed and select "Copy".

jonnew:
System.ArgumentException
  HResult=0x80070057
  Message=A device with the same name 'NeuropixelsV2eBetaHeadstage/NeuropixelsV2eBeta' has already been configured.
Parameter name: name
  Source=OpenEphys.Onix
  StackTrace:
   at OpenEphys.Onix.DeviceManager.RegisterDevice(String name, DeviceInfo deviceInfo) in C:\Users\open-ephys\public\Bonsai-Stuff\onix-refactor\OpenEphys.Onix\OpenEphys.Onix\DeviceManager.cs:line 33
   at OpenEphys.Onix.ConfigureNeuropixelsV2eBeta.<>c__DisplayClass25_0.<Process>b__0(ContextTask context) in C:\Users\open-ephys\public\Bonsai-Stuff\onix-refactor\OpenEphys.Onix\OpenEphys.Onix\ConfigureNeuropixelsV2eBeta.cs:line 118
   at OpenEphys.Onix.ObservableExtensions.<>c__DisplayClass4_1.<ConfigureContext>b__2(ContextTask ctx) in C:\Users\open-ephys\public\Bonsai-Stuff\onix-refactor\OpenEphys.Onix\OpenEphys.Onix\ObservableExtensions.cs:line 59
   at OpenEphys.Onix.ContextTask.ConfigureResources(StackDisposable disposable, Func`2 action) in C:\Users\open-ephys\public\Bonsai-Stuff\onix-refactor\OpenEphys.Onix\OpenEphys.Onix\ContextTask.cs:line 171
   at OpenEphys.Onix.ContextTask.ConfigureContext() in C:\Users\open-ephys\public\Bonsai-Stuff\onix-refactor\OpenEphys.Onix\OpenEphys.Onix\ContextTask.cs:line 152
   at OpenEphys.Onix.ContextTask.StartAsync(Int32 blockReadSize, Int32 blockWriteSize, CancellationToken cancellationToken) in C:\Users\open-ephys\public\Bonsai-Stuff\onix-refactor\OpenEphys.Onix\OpenEphys.Onix\ContextTask.cs:line 191
   at OpenEphys.Onix.StartAcquisition.<>c__DisplayClass8_0.<Process>b__1(IObserver`1 observer, CancellationToken cancellationToken) in C:\Users\open-ephys\public\Bonsai-Stuff\onix-refactor\OpenEphys.Onix\OpenEphys.Onix\StartAcquisition.cs:line 29
   at System.Reactive.Linq.QueryLanguage.<>c__DisplayClass1ed`1.<Create>b__1ec(IObserver`1 observer)
   at System.Reactive.AnonymousObservable`1.SubscribeCore(IObserver`1 observer)
   at System.Reactive.ObservableBase`1.Subscribe(IObserver`1 observer)

  This exception was originally thrown at this call stack:
    OpenEphys.Onix.DeviceManager.RegisterDevice(string, OpenEphys.Onix.DeviceInfo) in DeviceManager.cs
    OpenEphys.Onix.ConfigureNeuropixelsV2eBeta.Process.AnonymousMethod__0(OpenEphys.Onix.ContextTask) in ConfigureNeuropixelsV2eBeta.cs
    OpenEphys.Onix.ObservableExtensions.ConfigureContext.AnonymousMethod__2(OpenEphys.Onix.ContextTask) in ObservableExtensions.cs
    OpenEphys.Onix.ContextTask.ConfigureResources(OpenEphys.Onix.StackDisposable, System.Func<OpenEphys.Onix.ContextTask, System.IDisposable>) in ContextTask.cs
    OpenEphys.Onix.ContextTask.ConfigureContext() in ContextTask.cs
    OpenEphys.Onix.ContextTask.StartAsync(int, int, System.Threading.CancellationToken) in ContextTask.cs
    OpenEphys.Onix.StartAcquisition.Process.AnonymousMethod__1(System.IObserver<oni.Frame>, System.Threading.CancellationToken) in StartAcquisition.cs
    [External Code]

PR #134, opened 2024-07-17
Add optical stimulation support to headstage-64

jonnew:
- Update electrical stimulation support to maintain consistency
- Need to figure out how and when to perform valid DC current check

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64ElectricalStimulator.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64ElectricalStimulator.cs
@@ -26,6 +26,10 @@ namespace OpenEphys.Onix
     {
         public const int ID = 4;
 
+        // NB: could be read from REZ but these are costant
+        public const double DacBitDepth = 16;
+        public const double AbsMaxMicroAmps = 2500;
+
         // managed registers
         public const uint NULLPARM = 0; // No command
         public const uint BIPHASIC = 1; // Biphasic pulse (0 = monophasic, 1 = biphasic; NB: currently ignored)
--- a/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64ElectricalStimulator.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64ElectricalStimulator.cs
@@ -26,7 +26,7 @@ namespace OpenEphys.Onix
     {
         public const int ID = 4;
 
-        // NB: could be read from REZ but these are costant
+        // NB: could be read from REZ but these are constant
         public const double DacBitDepth = 16;
         public const double AbsMaxMicroAmps = 2500;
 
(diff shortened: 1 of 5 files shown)


OpenEphys.Onix/OpenEphys.Onix/Headstage64OpticalStimulatorTrigger.cs
```diff
@@ -0,0 +1,225 @@
...
+        [Precision(3, 1)]
+        public double Delay
+        {
+            get => delay.Value;
+            set => delay.OnNext(value);
+        }
+
+        // TODO: Should this be checked before TRIGGER is written to below and an error thrown if
+        // DC current is too high? Or, should settings be forced too keep DC current under some value?
+        [Description("Direct current required during burst (mA). Should be less than 50 mA.")]
```
jonnew, line 131: Right now, this is a read-only report of the DC current draw of the circuit when its operating. Ideally we either want to throw an error when its too high and the user tries to trigger or to force the parameters into bounds. Not sure what is the best option.

glopesdev: My main worry is that both approaches can leave the person baffled: one because the workflow crashes from out of nowhere; the other because the change didn't have the expected effect.

This kind of validation is a good example of the advantages of having an external `Configuration` object that the sink node receives. All validation happens at the moment of constructing the configuration object, rather than crashing the sink. You can also have a graphical GUI that validates your entire stim config when you click submit but before actually trying to send to the node.

The only thing this would preclude is the possibility of dynamically changing only specific individual properties and immediately writing to registers, but this is something I am still not sure is ever really required, especially if this kind of validation is warranted.

OpenEphys.Onix/OpenEphys.Onix/ConfigureHeadstage64ElectricalStimulator.cs
```diff
@@ -26,6 +26,10 @@ static class Headstage64ElectricalStimulator
     {
         public const int ID = 4;
 
+        // NB: could be read from REZ but these are costant
```
glopesdev, line 29:
```suggestion
        // NB: could be read from REZ but these are constant
```

Small typo. I guess these bits are a minor refactoring of the Headstage64 electrical stim. Usually these would go in a separate PR, but in the interests of time we can go with it.

OpenEphys.Onix/OpenEphys.Onix/Headstage64OpticalStimulatorTrigger.cs
jonnew: - GUI: Yes, this planned.
- Config Object: real-time manipulation by upstream workflow logic then because a huge pain because inner properties cannot be exposed as reactive elements. What you state about changing individual properties is probably not correct: e.g. one of them is the frequency of stimulation. This is a pretty obvious target for real-time control.

For now, I'm just going to leave as is and GUI will use to validate initial parameters. Real-time changes can result in large current draws and this might affect function I guess.

glopesdev:
@jonnew I may have to do a force push to remove my name from the commit message if you don't mind, but don't worry it will all end up there.

PR #147, opened 2024-07-24
Refactor common data frame properties into abstract base type

### LINKED ISSUE #140: ManagedFrame.FrameClock should be ManagedFrame.Clock
https://github.com/neurogears/onix-refactor/blob/260149bdfb441c8d49a606bbfd303728cbc42bd0/OpenEphys.Onix/OpenEphys.Onix/ManagedFrame.cs#L18C26-L18C27

Can I make this change and push to main?

glopesdev:
All ONI frames have a common structure which is used by all device data streams. This structure is currently replicated in the various `*DataFrame` types defined in the library. This generates redundancy and complicates both documentation and refactoring, since changes to this common signature are not tied together.

This PR introduces two new abstract base classes: `DataFrame` and `BufferedDataFrame` to help provide common properties for classes exposing data from either a single frame, or data buffered from multiple frames, respectively.

The `Heartbeat` device data node has been refactored to define a proper payload type and `HeartbeatDataFrame` and the legacy `ManagedFrame` class has been removed from the library as it is now unused.

Fixes #140


jonnew:
Yeah, this looks good.  I don't really see a downside. I did a quick test with headstage-64 and it seems to work well. Not sure if you have specific tests in mind you want me to run.

PR #151, opened 2024-07-24
Remove Rhd2164AmplifierDataFormat

jonnew:
- This property affected the raw data fomat produced by the RHD2164 chip. However, when twos-complement was selected, it was not respected by the Rhd2164Data node.
- Further, the ability to select between twos-comp and offset binary is not useful. More useful would be adding the ability to represent data as uV in the Rhd2164Data node.


glopesdev:
Sounds good, I would remove in that case the entire enum `Rhd2164AmplifierDataFormat` since it is now unused. We can always reintroduce it with new options when we bake in voltage conversion in a future release.

glopesdev (approved):

Issue #152, opened 2024-07-24
TS4231DataFrame is too low level to be useful


jonnew:
`TS4231Data` produces a stream of `TS4231DataFrame` objects

https://github.com/neurogears/onix-refactor/blob/f42fb50cfc9d845e806eecc74564b9b79c453418/OpenEphys.Onix/OpenEphys.Onix/TS4231DataFrame.cs#L5

these require a lot of post-processing to be useful. We need to include functionality for converting these data into 3D positions. The question is where to do that. Do people really care about the raw data format of this device or do they want get 3D position data directly? I would strongly argue that the later is the case and that internal buffering and conversion should be performed in order to produce pre-calculated 3D positions that are produced by this node rather than the current form of `TS4231DataFrame`.

If we want to do this conversion internally, then the relevant transform is provided here: https://github.com/open-ephys/Bonsai.ONIX/blob/main/Bonsai.ONIX/TS4231V1FrameToPosition.cs

I would suggest the modified DataFrame should look something like:

```csharp
public class TS4231DataFrame
{
    public ulong Clock { get; }

    public ulong HubClock { get; }

    public int SensorIndex { get; }

    public Vector3 Position {get};
}
```

Since each receiver can produces a 3D position, another question is if `TS4231Data` should produce data from only one sensor (in the way that `NeuropixelsData` nodes produce data from a single probe) or if a Frame should be produced when enough data has been collected from a given sensor to calcuate a position and the Sensor index included in the frame to indicate which photodiode it corresponds to. My vote is for the later.

glopesdev:
While I agree that doing basic 3D reconstruction over pulses detected from a single sensor is definitely useful and should be included in the library, I don't think baking this directly into the `TS4231Data` node is the right approach.

The reason is there are multiple possible approaches to do 3D reconstruction and pose estimation using these sensors, either based on fusion of multiple sensors, or multiple base stations, or even model-based approaches. By restricting access to the raw data we are excluding these possibilities a priori.

My preferred design here would be to introduce a new downstream node that takes care of both queueing and reconstructing the 3D poses, with its own data frame emitting values at a lower rate. This is similar to the post-processing nodes that handle the BNO data.

Finally, this would also make it easier for us to have multiple versions of the downstream node without breaking the `TS4231Data` stream.

jonnew:
Yes, these are good points.

[...]

PR #154, opened 2024-07-24
Decode Ts4231 data into 3D positions

jonnew:
- Attempt at addressing #152
- Appears to "work" but will need more testing to prove its fully functional
- Looking for feedback on the concept and implementation

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/TS4231Data.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/TS4231Data.cs
@@ -1,27 +1,46 @@
 ﻿using System;
 using System.ComponentModel;
 using System.Linq;
+using System.Reactive;
 using System.Reactive.Linq;
 using Bonsai;
+using OpenCV.Net;
 
 namespace OpenEphys.Onix
 {
-    public class TS4231Data : Source<TS4231DataFrame>
+    public class Ts4231Data : Source<Ts4231DataFrame>
     {
         [TypeConverter(typeof(TS4231.NameConverter))]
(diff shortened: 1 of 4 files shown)


jonnew:
For some reason, windows ignored my changes to the class names which ia attempted to make Camel case like the rest of the library. TS4231 -> Ts4231.

glopesdev:
@jonnew Actually according to the capitalization conventions in framework design guidelines, the upper-case spelling is correct in this case, since TS is a two-letter acronym: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers

OpenEphys.Onix/OpenEphys.Onix/TS4231Data.cs
```diff
@@ -1,27 +1,46 @@
...
 using System.Linq;
+using System.Reactive;
 using System.Reactive.Linq;
 using Bonsai;
+using OpenCV.Net;
 
 namespace OpenEphys.Onix
 {
-    public class TS4231Data : Source<TS4231DataFrame>
+    public class Ts4231Data : Source<Ts4231DataFrame>
```
glopesdev, line 11: As per comment above I would leave the names of these classes and files as they were. I named them in upper case since it follows the capitalization conventions of the .NET framework.

jonnew: Man I dont like that at all but sure.

jonnew:
I'm addressing this in the issue-152-attempt2 branch.

PR #155, opened 2024-07-24
Refactor device manager to avoid reference counting

glopesdev:
The current `DeviceManager` API is built on a common manager pattern used by other packages where nodes reserve a shared resource and manage disposal via reference counting. While this does work, it has several disadvantages which complicate existing code and usage of the interface:

1. Any node can reserve a device resource subject. This means that data nodes can effectively reference devices which do not exist in the configuration chain and no error will be reported. The data node simply is left hanging waiting for the device to be configured (which it never is since it doesn't exist).

2. Disposal logic is harder to follow since multiple nodes can trigger disposal of devices from the device manager, i.e. the data nodes could hold a device "hostage" in the device table by not releasing its reference, which would prevent the configuration chain from resetting and reconfiguring itself (leading to duplicate name entries).

3. Data node source code is more convoluted than it needs to be, since device reservation requires a disposable object which needs to be wrapped in a call to `Observable.Using` and unwrapped to finally arrive at the `DeviceInfo`. While this forms a repetitive pattern, it is error prone and hard to read.

Given the following, this PR proposes to break the symmetry of device reservation into producers and consumers.

Producers are restricted to configuration nodes, which register devices in the `DeviceManager` and manage a disposable to withdraw the reservation once they are disposed (i.e. acquisition is stopped, configuration is cancelled, etc).

Consumers are data nodes or other operators who simply lookup devices in the `DeviceManager` and throw if they do not exist. They do not pre-reserve entries based on the specified name and do not lie around waiting for hypothetical devices to show up.

This change drastically simplifies the device manager API, especially on the data node side, where `GetDevice` now directly returns the `DeviceInfo` object and makes behavior more consistent and less error-prone, since name mismatches will be flagged as errors rather than simply leaving the data node apparently hanging.


OpenEphys.Onix/OpenEphys.Onix/Bno055Data.cs
```diff
@@ -13,15 +13,13 @@ public class Bno055Data : Source<Bno055DataFrame>
...
-            return Observable.Using(
-                () => DeviceManager.ReserveDevice(DeviceName),
-                disposable => disposable.Subject.SelectMany(deviceInfo =>
-                {
-                    var device = deviceInfo.GetDeviceContext(typeof(Bno055));
-                    return deviceInfo.Context.FrameReceived
-                        .Where(frame => frame.DeviceAddress == device.Address)
-                        .Select(frame => new Bno055DataFrame(frame));
-                }));
+            return DeviceManager.GetDevice(DeviceName).SelectMany(deviceInfo =>
```
jonnew, line 16: Nice.

jonnew:
I dont really see a downside to this. The reasoning makes a lot sense to me. Quick tests i performed:

- Stream from headstage-64 and turn off power midstream. Workflow does not error (in this case this is OK because there is no indication of an issue other than via checking the FmcLinkController's lock status, which we have not implemented; that's another issue though!)
- stream data from npx-2.0 beta and turn off power midstream. All the good behavior of the issue-121/131 merge are still there.

If there is anything else you would like me to do let me know. I'm not sure I'm in a position to give a beneficial code review.

PR #166, opened 2024-07-26
Add TS4231 data source that automatically calculates 3D positions

jonnew:
Addresses #152

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/TS4231V1Data.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/TS4231V1Data.cs
@@ -6,19 +6,19 @@ using Bonsai;
 
 namespace OpenEphys.Onix
 {
-    public class TS4231Data : Source<TS4231DataFrame>
+    public class TS4231V1Data : Source<TS4231V1DataFrame>
     {
-        [TypeConverter(typeof(TS4231.NameConverter))]
+        [TypeConverter(typeof(TS4231V1.NameConverter))]
         public string DeviceName { get; set; }
 
-        public override IObservable<TS4231DataFrame> Generate()
+        public override IObservable<TS4231V1DataFrame> Generate()
         {
             return DeviceManager.GetDevice(DeviceName).SelectMany(deviceInfo =>
             {
-                var device = deviceInfo.GetDeviceContext(typeof(TS4231));
+                var device = deviceInfo.GetDeviceContext(typeof(TS4231V1));
                 return deviceInfo.Context.FrameReceived
                     .Where(frame => frame.DeviceAddress == device.Address)
-                    .Select(frame => new TS4231DataFrame(frame));
+                    .Select(frame => new TS4231V1DataFrame(frame));
             });
         }
     }
--- a/OpenEphys.Onix/OpenEphys.Onix/TS4231V1Data.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/TS4231V1Data.cs
@@ -16,10 +16,10 @@ namespace OpenEphys.Onix
             return DeviceManager.GetDevice(DeviceName).SelectMany(deviceInfo =>
             {
                 var device = deviceInfo.GetDeviceContext(typeof(TS4231V1));
-                var hubClockPerioduSec = 1e6 / device.Hub.ClockHz;
+                var hubClockPeriod = 1e6 / device.Hub.ClockHz;
                 return deviceInfo.Context.FrameReceived
                     .Where(frame => frame.DeviceAddress == device.Address)
-                    .Select(frame => new TS4231V1DataFrame(frame, hubClockPerioduSec));
+                    .Select(frame => new TS4231V1DataFrame(frame, hubClockPeriod));
             });
         }
     }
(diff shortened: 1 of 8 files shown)


OpenEphys.Onix/OpenEphys.Onix/TS4231V1Data.cs
```diff
@@ -6,19 +6,20 @@
...
         public string DeviceName { get; set; }
 
-        public override IObservable<TS4231DataFrame> Generate()
+        public override IObservable<TS4231V1DataFrame> Generate()
         {
             return DeviceManager.GetDevice(DeviceName).SelectMany(deviceInfo =>
             {
-                var device = deviceInfo.GetDeviceContext(typeof(TS4231));
+                var device = deviceInfo.GetDeviceContext(typeof(TS4231V1));
+                var hubClockPerioduSec = 1e6 / device.Hub.ClockHz;
```
glopesdev, line 19: For naming consistency I would remove the uSec suffix from variable and parameter names.

OpenEphys.Onix/OpenEphys.Onix/TS4231V1DataFrame.cs
```diff
@@ -1,24 +1,26 @@
...
+using System.Runtime.InteropServices;
+using oni;
 
 namespace OpenEphys.Onix
 {
-    public class TS4231DataFrame : DataFrame
+    public class TS4231V1DataFrame : DataFrame
     {
-        public unsafe TS4231DataFrame(oni.Frame frame)
+        public unsafe TS4231V1DataFrame(oni.Frame frame, double hubClockPerioduSec)
```
glopesdev, line 9: Same as above

OpenEphys.Onix/OpenEphys.Onix/TS4231V1GeometricPositionConverter.cs
```diff
@@ -0,0 +1,169 @@
+﻿using OpenCV.Net;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Numerics;
+using System.Reactive.Linq;
+
+namespace OpenEphys.Onix
+{
+    class TS4231PulseQueue
```
glopesdev, line 10: Just nitpick since class is internal but does this need the V1 suffix as well?

OpenEphys.Onix/OpenEphys.Onix/TS4231V1GeometricPositionConverter.cs
```diff
@@ -0,0 +1,169 @@
...
+            false, false, false, false,
+            false, false, false, true, // axis 0, station 1
+            false, true,  true,  false,
+            false, false, true,  false,
+            false, false, false, true  // axis 1, station 1
+        };
+
+        Dictionary<int, TS4231PulseQueue> PulseQueues = new();
+
+        public TS4231V1GeometricPositionConverter(uint hubClockFrequencyHz, Point3d baseSation1Origin, Point3d baseSation2Origin)
```
glopesdev, line 49: Typo on baseStation

OpenEphys.Onix/OpenEphys.Onix/TS4231V1GeometricPositionConverter.cs
```diff
@@ -0,0 +1,169 @@
...
+                return null;
+            }
+
+            // position measurement time is defined to be the mean of the data used
+            var time = queues.PulseTimes.ToArray();
+            var width = queues.PulseWidths.ToArray();
+
+            var t11 = time[2] + width[2] / 2 - time[0];
+            var t21 = time[5] + width[5] / 2 - time[3];
+            var theta0 = 2 * Math.PI * SweepFrequencyHz * t11 - Math.PI / 2;
```
glopesdev, line 110: Just a thought here: It could be useful to output these angles in the data frame, as they could be potentially used in other tracking algorithms for fusion and would spare the complication of calculating them.

glopesdev (commented):
Looks good, just a few typos. No need to address the other comments now.

OpenEphys.Onix/OpenEphys.Onix/TS4231V1DataFrame.cs
jonnew: But there is nothing constraining the units of the argument though, especially since this is an internal and therefore undocumented constructor

OpenEphys.Onix/OpenEphys.Onix/TS4231V1GeometricPositionConverter.cs
jonnew: Yah

OpenEphys.Onix/OpenEphys.Onix/TS4231V1GeometricPositionConverter.cs
jonnew: I think there is a lot of low level access one would potentially want and this is just one of those items. 

I like the solution now because the other data class provides as low level as you would ever want and this one provides the opposite. I dont think we should mix the two personally.

OpenEphys.Onix/OpenEphys.Onix/TS4231V1DataFrame.cs
jonnew: Im wrong, nevermind.

PR #171, opened 2024-07-27
Remove geometric prefix

glopesdev:
To keep naming simple we are omitting this prefix for all `TS4231V1` position streaming types. In fact, given its structure, `TS4231V1PositionDataFrame` can probably even be reused as-is for other methods in the future, since it simply represents a position of a given sensor relative to a base station, which in itself is method-agnostic.


glopesdev:
@jonnew As we discussed I am merging this one in advance of preparing the packaging PR. If any issues we can always revisit before release.

jonnew:
perfect, thank you.

PR #175, opened 2024-07-29
Internalize implementation specific members

### LINKED ISSUE #174: Potentially increase protection level of ContextTask API elements
I'm documenting the public API. As I do this, Im using this issue to document situations where I think there is potential for increasing protection level of API elements.

The first I found is `ContextTask.Reset()`

https://github.com/neurogears/onix-refactor/blob/fb191f08a8377a189614541c6f648c2f2669457b/OpenEphys.Onix/OpenEphys.Onix/ContextTask.cs#L77

which only referenced by a function in the same class. It could be that we want external elements to force a reset in the future, e.g. in a GUI, but right now prototype GUIs have not performed hardware interaction.

glopesdev:
To increase degrees of freedom in maintaining future versions of the package we are internalizing several implementation specific members which are not required by the public surface API and are potential targets for future changes and refactor.

Fixes #174

### DIFF
--- a/OpenEphys.Onix/OpenEphys.Onix/HubDeviceFactory.cs
+++ b/OpenEphys.Onix/OpenEphys.Onix/HubDeviceFactory.cs
@@ -25,7 +25,7 @@ namespace OpenEphys.Onix
             }
         }
 
-        protected string GetFullDeviceName(string deviceName)
+        internal string GetFullDeviceName(string deviceName)
         {
             return !string.IsNullOrEmpty(_name) ? $"{_name}/{deviceName}" : string.Empty;
         }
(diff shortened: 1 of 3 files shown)


OpenEphys.Onix/OpenEphys.Onix/HubDeviceFactory.cs
```diff
@@ -25,7 +25,7 @@ public string Name
             }
         }
 
-        protected string GetFullDeviceName(string deviceName)
```
jonnew, line 28: This is the only one that surprised me a bit since this is a decrease in protection as far as I understand

glopesdev: @jonnew Its actually not since `protected` applies to classes outside the assembly, i.e. anyone inheriting from the class would be able to use the method.

glopesdev: And interestingly "protected internal" means the union of access qualifiers rather than the intersection so it wouldn't help either unfortunately.

PR #189, opened 2024-08-02
Minor updates to documentation

jonnew:
- Clarification and typo fixes

### DIFF
--- a/OpenEphys.Onix1/MultiDeviceFactory.cs
+++ b/OpenEphys.Onix1/MultiDeviceFactory.cs
@@ -6,19 +6,26 @@ namespace OpenEphys.Onix1
 {
     /// <summary>
     /// Provides an abstract base class for configuration operators responsible for
-    /// registering all devices in an ONI device aggregate in the context device table.
+    /// registering logical groups of <see cref="oni.Device"/>s.
     /// </summary>
     /// <remarks>
     /// <para>
-    /// ONI devices are often grouped into multi-device aggregates connected to hubs or
-    /// headstages. These aggregates provide access to multiple devices through hub-specific
-    /// addresses and usually require a specific sequence of configuration steps to determine
-    /// operational port voltages and other link-specific settings.
+    /// The ONI standard states that devices are grouped into aggregates called hubs, each of which is
+    /// governed by a single, potentially asynchronous clock and share a common base address. The devices on
+    /// a headstage are an example of a hub. Devices within a hub are accessed through hub-specific addresses
+    /// and usually require a specific sequence of configuration steps prior to acquisition.
     /// </para>
     /// <para>
-    /// These multi-device aggregates are the most common starting point for configuration
-    /// of an ONI system, and the <see cref="MultiDeviceFactory"/> provides a modular abstraction
-    /// for flexible assembly and sequencing of multiple such aggregates.
+    /// This class allows configuration of logical device groups of <see cref="oni.Device"/>s across ONI-defined
+    /// hubs. For instance, the group of devices within a headstage (a single hub) can be combined with a device
+    /// from another hub that is used to control its port voltage and communication status
+    /// (e.g. <see cref="ConfigureHeadstage64"/>). Alternatively, diagnostic devices that are present within
+    /// an ONI hub can be omitted from a device group to aid it useability (e.g. <see cref="ConfigureBreakoutBoard"/>).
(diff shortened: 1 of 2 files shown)


OpenEphys.Onix1/MultiDeviceFactory.cs
```diff
@@ -6,19 +6,26 @@ namespace OpenEphys.Onix1
...
     /// </para>
     /// <para>
-    /// These multi-device aggregates are the most common starting point for configuration
-    /// of an ONI system, and the <see cref="MultiDeviceFactory"/> provides a modular abstraction
-    /// for flexible assembly and sequencing of multiple such aggregates.
+    /// This class allows configuration of logical device groups of <see cref="oni.Device"/>s across ONI-defined
+    /// hubs. For instance, the group of devices within a headstage (a single hub) can be combined with a device
+    /// from another hub that is used to control its port voltage and communication status
+    /// (e.g. <see cref="ConfigureHeadstage64"/>). Alternatively, diagnostic devices that are present within
+    /// an ONI hub can be omitted from a device group to aid it useability (e.g. <see cref="ConfigureBreakoutBoard"/>).
```
bparks13, line 23:
```suggestion
    /// an ONI hub can be omitted from a device group to aid its useability (e.g. <see cref="ConfigureBreakoutBoard"/>).
```

bparks13 (changes requested):
Looks good, I'll go ahead and implement the change listed and then we can merge the PR

bparks13 (approved):

OpenEphys.Onix1/MultiDeviceFactory.cs
```diff
@@ -6,19 +6,26 @@ namespace OpenEphys.Onix1
 {
     /// <summary>
     /// Provides an abstract base class for configuration operators responsible for
-    /// registering all devices in an ONI device aggregate in the context device table.
+    /// registering logical groups of <see cref="oni.Device"/>s.
```
glopesdev, line 9: This somehow seems now less clear, since it makes it seem the operator is registering the group rather than each individual device. Maybe we shouldn't talk about "device table" since that is a hardware specific concept, but mentioning that there is a context "device manager" might be important so people understand why sometimes the drop down shows devices and sometimes not.

OpenEphys.Onix1/MultiDeviceFactory.cs
```diff
@@ -6,19 +6,26 @@ namespace OpenEphys.Onix1
...
+    /// The ONI standard states that devices are grouped into aggregates called hubs, each of which is
+    /// governed by a single, potentially asynchronous clock and share a common base address. The devices on
+    /// a headstage are an example of a hub. Devices within a hub are accessed through hub-specific addresses
+    /// and usually require a specific sequence of configuration steps prior to acquisition.
     /// </para>
     /// <para>
-    /// These multi-device aggregates are the most common starting point for configuration
-    /// of an ONI system, and the <see cref="MultiDeviceFactory"/> provides a modular abstraction
-    /// for flexible assembly and sequencing of multiple such aggregates.
+    /// This class allows configuration of logical device groups of <see cref="oni.Device"/>s across ONI-defined
```
glopesdev, line 19:
```suggestion
    /// This class allows configuration of logical groups of <see cref="oni.Device"/>s across ONI-defined
```

OpenEphys.Onix1/MultiDeviceFactory.cs
```diff
@@ -34,10 +41,10 @@ internal MultiDeviceFactory()
         }
 
         /// <summary>
-        /// Gets or sets a unique hub device name.
+        /// Gets or sets a unique device group name.
```
glopesdev, line 44: Could be useful to clarify here that this name is used as a prefix for individual device names in the group.

OpenEphys.Onix1/MultiDeviceFactory.cs
jonnew: Really, as far as I see, we are both wrong. 
- Device table is hardware-agnostic, its required by the ONI API regardless of hardware
- This class does not configure aggregates of oni.Devices (nullifying impact of last bullet), but OpenEphys.Onix1's concept of device, which is much looser (e.g. any dev using passthrough)

I m going to remove references to oni.Devices because its just not correct. I agree with Device manager comment ill try to add some text.

PR #190, opened 2024-08-03
Fix register mapping for Neuropixels 2.0 public probes

jonnew:
- Neuropixels 2.0 has a completely different register map compared to 2.0-beta. This is reflected in the changes in this commit.
- Common static helper functions for 1.0, 2.0, and 2.0-beta were moved to BitHelper.cs and NeuropixelV2.cs


glopesdev (approved):
@jonnew This all looks great, were you able to get shift register OK checks with these changes?

jonnew:
Its hard to know with 100 percent certainty, due the lack of proper Neuropixels documentation in general and unavailability of their API (even though its supposed to be open source given the terms of IMECs contract) that everything is functional but course tests with Jakob reveal that we are definitely shifting electrode positions etc.

jonnew:
Side note: I changed the SR check reporting level from raising an exception to printing to the Error stream. The reason is because on multishank probes the users should be able to use a probe even if one shank is broken.

glopesdev:
I think this is totally reasonable and appropriate, especially given the current state of the integration, we can update this once we get certainty.

jonnew:
Yeah to be honest there are a lot of systematic issues I would like to go over that relate to shared state between devices that use passthrough and the way that gain corrections are handled here, which I think is pretty bad: declare as nullable double, use if sn detected and cast to double just feels gross.

jvoigts:
I'll have time to debug this some more next week. I dont think I'm seeing any SR check errors but I have not verified that the channel map is behaving as expected.

jonnew:
@jvoigts When you use the library with np2.0 probes, please let me know so I can verify the SR programming is working for the whole probe

PR #193, opened 2024-08-04
Merge 88 into 186

jonnew:
- I don't think this should be merged as is. 
- There are issues with how the GUI is interacting with the main library
- There are stability issues with the GUI
- We need to do a code review and think about what we want to take and leave from this PR


bparks13:
Now that `NeuropixelsV2e` dialogs are merged, do we need anything from this draft PR?

PR #196, opened 2024-08-06
Publish packages to github on release

glopesdev:
This updates the release automation scripts to publish packages to github without pre-release suffixes on creating a new release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,17 @@ env:
   CiRunPushSuffix: ${{ github.ref_name }}-ci${{ github.run_number }}
   CiRunPullSuffix: pull-${{ github.event.number }}-ci${{ github.run_number }}
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      build-suffix: ${{ steps.setup-build.outputs.build-suffix }}
+    steps:
+      - name: Setup Build
+        id: setup-build
+        run: echo "build-suffix=${{ github.event_name == 'push' && env.CiRunPushSuffix || github.event_name == 'pull_request' && env.CiRunPullSuffix || null }}" >> "$GITHUB_OUTPUT"
+
   build:
+    needs: [setup]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -60,7 +71,7 @@ jobs:
     permissions:
       packages: write
     needs: [build]
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' || github.event_name == 'release'
```
bparks13, line 74: Based on our discussion, adding some comments here. Some of this was pulled from https://github.com/actions/runner/issues/491 for reference.

The combination of `needs` and `if` appears to be inconsistent with what we want to happen. By specifying `needs: [build]` this job will not occur until after the build job runs, but it does not guarantee that the build job was successful. In the example given [here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs) by specifying `if`, whenever the condition supplied is true this step will run whether or not the jobs specified in `needs` were successful.

To check if a job was successful, we can add `needs.build.result == 'success'` to ensure that the build process succeeded.

This might all be a moot point because the nuget push command should fail anyway if there are no packages to push.

glopesdev: @bparks13 thanks for the reference issue and the comments, this is really properly confusing but I think after going through a lot of the issue comments my feeling is that the `if` clause does inject an implicit `success()` check on top of the condition. The example you referenced is actually of a job where you do not want to do that implicit check so they add an explicit `always()` in the body of `if`.

My understanding is that the body of  whole issue is actually complaining that github actions should not do the explicit check and make `needs` be just a sequencer, but it looks like they didn't change it and the issue was closed on Feb last year with no PR merged anywhere.

I am of the camp that `needs` should indeed imply `success()` by default, since that is the most sane default. Good to know that there is a way to make it so you can run regardless of success, but I don't think we actually need to do anything here.

Just to be sure, I also just tested it on my private fork of this repo and injected `if: false` on the build job and both jobs get skipped correctly. Adding `always() &&` the rest of the condition does cause the publish action to run in this condition, but not by default.

bparks13 (approved):

PR #213, opened 2024-08-13
Convert BreakoutAnalogOutput data to offset binary

### LINKED ISSUE #205: Incorrect data type for BreakoutAnalogOutput
The analog outputs on the breakout board requires unsigned 16-bit integer data that is interpreted by the hardware as offset binary:

https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/fmc-analog-io.html#host-to-device-data-frames

However, signed data is used throughout the output node resulting in incorrect values being sent to the hardware:

https://github.com/open-ephys/onix-bonsai-onix1/blob/7ac700fb8a696d4d9f8d97c9d1b85765b9951ca6/OpenEphys.Onix1/BreakoutAnalogOutput.cs#L100

This issue applies to both `Volts` and `S16` output options because they are both ultimately transformed to `S16` before being sent. 

image

jonnew:
- Fixes #205 
- The analog outputs on the breakout board expect 16-bit words in offset binary
- We were using twos-complement. This fix converts all outgoing data to offset binary before the final calls to device.Write()

### DIFF
--- a/OpenEphys.Onix1/BreakoutAnalogOutput.cs
+++ b/OpenEphys.Onix1/BreakoutAnalogOutput.cs
@@ -37,7 +37,7 @@ namespace OpenEphys.Onix1
         /// </summary>
         /// <param name="source"> A sequence of 12xN sample matrices containing the analog data to write to channels 0 to 11.</param>
         /// <returns> A sequence of 12xN sample matrices containing the analog data that were written to channels 0 to 11.</returns>
-        public override IObservable<Mat> Process(IObservable<Mat> source)
+        public override unsafe IObservable<Mat> Process(IObservable<Mat> source)
         {
             var dataType = DataType;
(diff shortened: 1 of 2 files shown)


bparks13:
To automatically close the existing issue, we can link it from the PR back to the issue with closing language. For example:
Fixes #205

OpenEphys.Onix1/BreakoutAnalogOutput.cs
```diff
@@ -108,6 +115,12 @@ public IObservable<short[]> Process(IObservable<short[]> source)
                 return source.Do(data =>
                 {
                     AssertChannelCount(data.Length);
+                    var samples = new ushort[data.Length];
```
jonnew, line 118: There is potentially an opencv operation that would vectorize this. `cv.xors` is what we should look into.

PR #214, opened 2024-08-13
Add configuration GUI for NeuropixelsV2e

### LINKED ISSUE #88: NeuropixelsV2 channel configuration GUI

### LINKED ISSUE #137: ConfigureNeuropixelsV2e node typos
image
Popup should say: Probe serial number xxx does not match calibration file serial number xxx, but it says calibraiton.

image
Popup should say: “Calibration file must be specified.”, but it says calibraiton.

image
Configuration Panel should say NeuropixelsV2e but it says NeuropixelsV2.

### LINKED ISSUE #182: Reduce aliasing effect in Neuropixels GUIs when fully zoomed out
In Neuropixels GUIs, when the probe is fully zoomed out there is some aliasing occurring where contacts do not appear to be the same size. This is probably an effect of using `ZedGraph`, but there might be some options that can reduce this effect. Look into `ZedGraph` drawing options as well as general WinForms controls options that could help with this.

bparks13:
This PR aims to add a configuration GUI at each level of `NeuropixelsV2e`. Individual `ProbeConfiguration` variables can be selected to open a GUI that modifies the individual probe configuration, or the `ConfigureNeuropixelsV2eHeadstage` can be double-clicked to open up a dialog with both possible probes in tabs, and the `NeuropixelsV2eBno055` in a separate tab.

After opening the GUI, there are two tabs that will initially be blank (Probe A / Probe B); to show the probes for editing, select a calibration file on the right side. Alternatively, you can set the calibration file from the property editor in the main Bonsai editor.

Controls: Mouse wheel zooms in/out, left-click and drag will select contacts, left-click will select a single contact or will clear the selection if no contact is selected, middle-click and drag will pan the contacts.

On the right side there is a scroll bar that can be used to move the probe up and down, which is useful when zooming in to get to the top or bottom quickly.

To change the enabled channels (enabled channels are in blue, disabled channels are gray), select any number of channels, and choose Enable Contacts. To clear the currently selected channels without enabling them, press Clear Selection.

Channel presets can also be used for each probe separately, choose any preset from the drop-down menu and it will automatically populate the correct channels.

Menu drop-downs (File -> NeuropixelsV2e -> Probe A/B) can be used to load or save the JSON file (the ProbeInterface file), or to reset the selected channels you can load the default from that menu.

Closes #88 
Fixes #137 
Partially fixes #182


jonnew:
The first thing I will do is look at the `ProbeInterface` code so we can create a nuget package. Once thats taken care of, we can remove the submodule and try, reference the nuget package instead, and try the automated build again.

jonnew:
Functionally, I have not found an issue yet. Design wise, here are my notes:

image

bparks13:
@jonnew  I'll get to work on these. Couple of clarifying comments/questions:

File menu drop-down: Are you wanting it to be a flat file menu, or modifying the current levels to match the strings given? Meaning, would it still be NeuropixelsV2e -> Probe A | B -> "new strings here"? That would be easy enough to do.

The documentation link is something that I saw in the Bonsai.ONIX.Design library and copied over, currently it is (or should be) pointing to the new Onix1 Bonsai docs page, but it does not automatically update. The F1 help that can be reached from the main Bonsai editor points to the same or similar page, so we should be fine just removing the documentation link label from the GUI.

jonnew:
I dont like `ParseGainCalibrationFile()` because it does not reflect hardware and does not check that the file is in the correct format. Its nice to have this information below, but what happens if there is nonsense in the file?

bparks13:
@jonnew This function is not meant to reflect hardware, it is purely a parsing function for the file itself because the `ReadGainCorrection()` method assumes that we already know the serial number, but in the GUI we have no way to access the serial number on the hardware. I added it as a sanity check while configuring the GUI so that the user could make sure that they were using the correct gain calibration file before trying to run the workflow and getting an exception then.

Regarding the file format, there is also no checking for the file format when the workflow is running and the gain correction is read, so what I can do is add some checking of the file format and also refactor it a little bit so that the two methods are calling the same underlying function to ensure consistency.

jonnew (approved):
Review was performed outside of github via integration testing and talking on discord.

jonnew:
Merged 🥳

PR #215, opened 2024-08-14
Improve description of neuropixels calibration files

### LINKED ISSUE #200: More information on ADC calibration files and gain calibration files for NP headstages
image
I think we need more information on where ADC calibration files and gain calibration files come from. They have to ask IMEC for these, and the file names must match the serial number of their Neuropixels probe. Perhaps with an admonition about how they shouldn't change those calibration file names?

jonnew:
- More explaination
- More emphasis on their importance
- Path to getting a new one if yours is lost
- Fixes #200

### DIFF
--- a/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
+++ b/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
@@ -90,8 +90,11 @@ namespace OpenEphys.Onix1
         /// Gets or sets the path to the gain calibration file.
         /// </summary>
         /// <remarks>
-        /// Each probe must be provided with a gain calibration file that contains calibration data
-        /// specific to each probe. This file is mandatory for accurate recordings.
+        /// Each probe is linked to a gain calibration file that contains a gain adjustments determined by IMEC during
+        /// factory testing. Electrode voltages are scaled using these values to ensure they can be accurately compared
+        /// across probes. Therefore, using the correct gain calibration file is mandatory to create standardized recordings.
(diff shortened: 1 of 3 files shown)


cjsha (approved):
### typo

"contains a gain adjustments". Change to "contains gain adjustments" or "contains a gain adjustment"

### Other comments/suggestions

The rest of this comment is nitpicks. I don't feel strongly about anything below so feel free to pick and choose what looks good to you and merge.

Though it's sorta implied in your text already, it might be worth explaining this explicitly:

> Calibration files are probe-specific and not interchangeable across probes. Calibration file names must contain the serial number of the corresponding probe and should not be changed from the factory-given name to avoid confusion.

You have:

> Electrode voltages are scaled using these values to ensure they can be accurately compared across probes. Therefore, using the correct gain calibration file is mandatory to create standardized recordings. 

I suggest something like:

> Using the correct calibration file is mandatory for Neuropixels to take calibrated measurements and be able to produce useful/meaningful data. 

^ I think my text above is more emphatic about setting the correct calibration file. However, it lacks some under-the-hood explanation that your text has.

My preference is to remove the stricken out text:

> If you have lost track of the gain calibration file for your probe, ~~you can~~ email IMEC at neuropixels.info@imec.be with the probe serial number to retrieve a new copy.

I didn't comment with line numbers bc these things I brought up repeat on a few different lines

jonnew:
Addressed these in the latest commit. Accepted most. Disregarded "useful/meaningful" vesus "standardized" for gain calibration because later is actually the case.

PR #216, opened 2024-08-14
Expose headstage port status data stream

### LINKED ISSUE #156: Use FMC Link Controller data stream
Aside from controlling port voltages, setting passthough vs normal mode, etc the [FMC Link Controller](https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/fmc-link-control.html) exposes a data stream that is useful for monitoring port state. 

https://open-ephys.github.io/onix-docs/Hardware%20Guide/Datasheets/fmc-link-control.html#device-to-host-data-frames

This provides information on e.g. the loss of lock or deserialization parity check failures. AFIAK, this stream is not currently used in the library. It would be a very nice thing to have access to in terms of a savable data stream that can be used both for diagnoistic reasons and for potentially taking action within the library to e.g. automatically attempt to re-establish lock.

Im not sure where such a device really belongs -- currently the register access portion of the device is used but hidden inside headstage aggregates. I really like this and want to preserve that. But it kinda conflicts with then creating a Data node that is linked to --- what?

jonnew:
- Fixes #156
- I feel that FmcLinkController  and ConfigureFmcLinkController should be renamed PortController and ConfigurePortController, respectively to match their implicity configurate via the "Port" property in headstages and miniscopes, as well as their labelling on the hardware

### DIFF
--- a/OpenEphys.Onix1/ConfigureNeuropixelsV2eLinkController.cs
+++ b/OpenEphys.Onix1/ConfigureNeuropixelsV2eLinkController.cs
@@ -29,7 +29,7 @@ namespace OpenEphys.Onix1
 
         void SetVoltage(DeviceContext device, double voltage)
         {
-            device.WriteRegister(FmcLinkController.PORTVOLTAGE, 0);
+            device.WriteRegister(PortController.PORTVOLTAGE, 0);
             Thread.Sleep(200);
             device.WriteRegister(FmcLinkController.PORTVOLTAGE, (uint)(10 * voltage));
             Thread.Sleep(200);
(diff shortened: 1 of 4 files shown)


OpenEphys.Onix1/ConfigureNeuropixelsV2eLinkController.cs
```diff
@@ -29,7 +29,7 @@ protected override bool ConfigurePortVoltage(DeviceContext device)
 
         void SetVoltage(DeviceContext device, double voltage)
         {
-            device.WriteRegister(FmcLinkController.PORTVOLTAGE, 0);
+            device.WriteRegister(PortController.PORTVOLTAGE, 0);
```
jonnew, line 32: This is a mistake. I tried to leave this change for the discussion and instead broke the code.

PR #217, opened 2024-08-15
Expose headstage port status data stream

### LINKED ISSUE #156: Use FMC Link Controller data stream

jonnew:
- Fixes #156
- Rename `*FmcLink*` to `*Port*`
- Improve error string when SERDES lock cannot be established
- Move PortName enum into ConfigurePortController.cs

### DIFF
--- a/OpenEphys.Onix1/ConfigureHeadstage64.cs
+++ b/OpenEphys.Onix1/ConfigureHeadstage64.cs
@@ -5,13 +5,13 @@ using System.Threading;
 namespace OpenEphys.Onix1
 {
     /// <summary>
-    /// A class that configures an ONIX headstage-64 in the specified port.
+    /// A class that configures an ONIX headstage-64 on the specified port.
     /// </summary>
     [Description("Configures an ONIX headstage-64 in the specified port.")]
     public class ConfigureHeadstage64 : MultiDeviceFactory
     {
         PortName port;
-        readonly ConfigureHeadstage64LinkController LinkController = new();
+        readonly ConfigureHeadstage64PortController LinkController = new();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ConfigureHeadstage64"/> class.
@@ -136,7 +136,7 @@ namespace OpenEphys.Onix1
             yield return OpticalStimulator;
         }
 
-        class ConfigureHeadstage64LinkController : ConfigureFmcLinkController
+        class ConfigureHeadstage64PortController : ConfigurePortController
         {
             protected override bool ConfigurePortVoltage(DeviceContext device)
             {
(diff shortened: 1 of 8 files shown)


OpenEphys.Onix1/ConfigureHeadstage64.cs
```diff
@@ -5,13 +5,13 @@
...
     /// <summary>
-    /// A class that configures an ONIX headstage-64 in the specified port.
+    /// A class that configures an ONIX headstage-64 on the specified port.
     /// </summary>
     [Description("Configures an ONIX headstage-64 in the specified port.")]
     public class ConfigureHeadstage64 : MultiDeviceFactory
     {
         PortName port;
-        readonly ConfigureHeadstage64LinkController LinkController = new();
+        readonly ConfigureHeadstage64PortController LinkController = new();
```
bparks13, line 14: Is it worth updating the name of the variable itself to reflect the class name change?

OpenEphys.Onix1/ConfigureNeuropixelsV1eHeadstage.cs
```diff
@@ -5,13 +5,13 @@
...
     /// <summary>
-    /// A class that configures a NeuropixelsV1e headstage.
+    /// A class that configures a NeuropixelsV1e headstage on the specified port.
     /// </summary>
     [Description("Configures a NeuropixelsV1e headstage.")]
     public class ConfigureNeuropixelsV1eHeadstage : MultiDeviceFactory
     {
         PortName port;
-        readonly ConfigureNeuropixelsV1eLinkController LinkController = new();
+        readonly ConfigureNeuropixelsV1ePortController LinkController = new();
```
bparks13, line 14: Same as above

bparks13 (approved):
Looks good on my end, just a comment on whether or not to change the variable names to match.

OpenEphys.Onix1/ConfigureHeadstage64.cs
jonnew: I considered this but `PortController` would conflict with the type of the same name. I was thinking about it and then gave up. If you have an idea for replacement ill take it.

PR #220, opened 2024-08-15
Implement ConfigureNeuropixelV1/2eBno055.Enable

### LINKED ISSUE #191: ConfigureNeuropixelsV2eBno055.Enable has no effect
This also is true of `ConfigureNeuropixelsV1eBno055.Enable `. Basically the property has no effect the `Process` method and is not attached to `DeviceInfo` for use in the Data node:

https://github.com/open-ephys/onix-bonsai-onix1/blob/7cb45a2af4fc387fcfbf0a2028c09b78ccb41c06/OpenEphys.Onix1/ConfigureNeuropixelsV2eBno055.cs#L45

This has a similar flavor to #165.

jonnew:
- Fixes #191
- Previously ConfigureNeuropixelV1/2eBno055.Enable was not doing anything
- Now enable state is passed through a custom DeviceInfo object to the NeuropixelV1/2eBno055Data operator. If its false, the operator returns an Observable.Empty<Bno055DataFrame>.

### DIFF
--- a/OpenEphys.Onix1/NeuropixelsV1eBno055Data.cs
+++ b/OpenEphys.Onix1/NeuropixelsV1eBno055Data.cs
@@ -39,32 +39,37 @@ namespace OpenEphys.Onix1
         public unsafe IObservable<Bno055DataFrame> Generate<TSource>(IObservable<TSource> source)
         {
             return DeviceManager.GetDevice(DeviceName).SelectMany(
-                deviceInfo => Observable.Create<Bno055DataFrame>(observer =>
+                deviceInfo =>
                 {
-                    var device = deviceInfo.GetDeviceContext(typeof(NeuropixelsV1eBno055));
-                    var passthrough = device.GetPassthroughDeviceContext(typeof(DS90UB9x));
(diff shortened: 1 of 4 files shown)


bparks13 (approved):
Looks good to me.

On a related note, I noticed that in `ConfigureBno055` it does not cache the Enable property. To keep things clean I can open a separate issue for that, or we can add it in here.

OpenEphys.Onix1/NeuropixelsV1eBno055Data.cs
```diff
@@ -39,32 +39,37 @@ public override IObservable<Bno055DataFrame> Generate()
...
+                                fixed (byte* dataPtr = data)
+                                {
+                                    frame = new Bno055DataFrame(clock, (Bno055DataPayload*)dataPtr);
+                                }
+                            });
+
+                            if (frame != null)
                             {
-                                frame = new Bno055DataFrame(clock, (Bno055DataPayload*)dataPtr);
+                                observer.OnNext(frame);
```
jonnew, line 68: This was a copy paste. Not sure if its wrong, need to investigate

PR #223, opened 2024-08-16
Add more informative titles to project files

bparks13:
Instead of using the namespace as the title for each project (i.e., `OpenEphys.Onix1`), give a more informative and user-friendly title (e.g., `Bonsai Onix1 - Open Ephys`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
-.vs/
+.vs*/
 /artifacts/
 .bonsai/Packages/
 .bonsai/*.exe
 .bonsai/*.exe.settings
 .bonsai/*.exe.WebView2/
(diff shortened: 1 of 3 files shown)


bparks13:
We have decided not to modify the titles, since most package managers do not actually use the title, but rather the ID. This includes in-development changes to the Bonsai package manager. Rather, the description should (and already does) specify that this is a Bonsai library.

Future work will include specifying the [PackageType](https://learn.microsoft.com/en-us/nuget/create-packages/set-package-type?tabs=dotnet) as `Dependency;BonsaiLibrary` to allow differentiation between regular NuGet packages and Bonsai-specific packages.

Issue #122, opened 2024-06-28
Abstract device combinations


jonnew:
Some ONI devices a best treated as groups at the level of this library. A straightforward example are the two RHS2116 chips on headstage-rhs2116. These chips are synchronously sampled and transmitted at the level of headstage hardware. They go to a single connector, which is generally plugged into one probe. They are effectively a single 32-channel chip. Yet in this library, they are treated separately. This is completely fine for detailed manual configuration, but by the time the user is putting `ConfigureHeadstageRhs2116` in their workflow, it would be best to expose only a single, virtual, 32-channel `ConfigureRhs2116Array`. Settings would be shared between the two chips using a bit of software muxing to select the underlaying `ConfigureRhs2116` objects. A `GroupedDeviceNameConverter` would be needed create names of applicable groups of devices. Data would be combined in a non-stupid version of the following (which actually works so long as both chips are enabled):

``` csharp
using System;
using System.ComponentModel;
using System.Linq;
using System.Reactive.Linq;
using Bonsai;
using OpenCV.Net;

namespace OpenEphys.Onix
{
    public class Rhs2116DualData : Source<Rhs2116DataFrame>
    {

        private Rhs2116Data rhsA { get; set; } = new Rhs2116Data();
        private Rhs2116Data rhsB { get; set; } = new Rhs2116Data();

        [TypeConverter(typeof(Rhs2116.NameConverter))]
        public string DeviceName 
        { 
            get 
            { 
                return rhsA.DeviceName.Remove(rhsA.DeviceName.Length - 1); 
            } 
            set 
            { 
                rhsA.DeviceName = value.Remove(value.Length - 1) + 'A';
                rhsB.DeviceName = value.Remove(value.Length - 1) + 'B';
            } 
        }

        public int BufferSize
        {
            get
            {
                return rhsA.BufferSize;
            }
            set
            {
                rhsA.BufferSize = value;
                rhsB.BufferSize = value;
            }
        }

        public unsafe override IObservable<Rhs2116DataFrame> Generate()
        {
            var a = rhsA.Generate();
            var b = rhsB.Generate();

            return a.Zip(b, (lhs, rhs) =>
            {
                var ampMat = new Mat(2 * lhs.AmplifierData.Rows, lhs.AmplifierData.Cols, lhs.AmplifierData.Depth, lhs.AmplifierData.Channels);
[...]

PR #224, opened 2024-08-16
Dual rhs2116 device

jonnew:
- Attempt at #122
- This is a draft because its non-functional for reasons unknow


jonnew:
Moved to

PR #226, opened 2024-08-20
Add Acquisition or Configuration category to public properties

bparks13:
To ensure a consistent user experience across all operators, this PR adds a string to the Category attribute for all public properties that defines the property as `Acquisition`, `Configuration`, or `Devices`.

`Configuration` properties are set prior to the start of the workflow. These are not updated in real-time, and are configured once when the workflow starts.

`Acquisition` properties can have an initial state defined before the workflow begins, but they can be updated in real-time with new values sent to hardware while the workflow is running.

`Devices` properties are only applicable for aggregates, where this category is applied to each device in an aggregate.


OpenEphys.Onix1/ConfigureNeuropixelsV2eLinkController.cs
```diff
@@ -4,7 +4,6 @@ namespace OpenEphys.Onix1
 {
     class ConfigureNeuropixelsV2eLinkController : ConfigureFmcLinkController
     {
-
```
jonnew, line 7: What changed here?

jonnew (changes requested):
Can you do a quick review of all these changes and make sure that they comport with my commands on `MutliDeviceFactory.Name`? Basically, a property should be marked if its something exposed to the user that changes the behavior of a device at start or start/runtime. Anything outside of that scope should not be marked.

Aside from this, in `ConfigureBreakoutAnalogIO` the direction properties are all marked as Acquisition, but should be configuration. Thats left over from when they were settable at runtime and I forgot to change.

OpenEphys.Onix1/ConfigureNeuropixelsV2eLinkController.cs
bparks13: There was an empty line that I removed, no functional changes

OpenEphys.Onix1/MultiDeviceFactory.cs
```diff
@@ -44,6 +44,7 @@ internal MultiDeviceFactory()
         /// the <see cref="SingleDeviceFactory.DeviceName"/> of each device in the the group. 
         /// </remarks>
         [Description("The unique device group name.")]
+        [Category(ConfigurationCategory)]
```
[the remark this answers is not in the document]
bparks13: I had modified these because they are shown to user from the headstage and they set the name for all subsequent devices before the workflow begins and are not modified in real-time. In my mind, the aggregate name is a Configuration property for that reason.

Plus, if we don't mark these with a category, they get lumped into a `Misc` category. So for consistency everything that is exposed to the user got a category attribute.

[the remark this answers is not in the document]
bparks13: re: marking the setter as private.

Tested this locally, it throws the following error when opening a workflow with an aggregate:
image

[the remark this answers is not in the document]
jonnew: ok, so I was wrong on all accounts :). Not the first time.

jonnew:
The remaining though i have before merging is the members of a `MultiDeviceFactory` being marked as `Configuration`

image

I understand the reasoning here, as the properties referred to by the attribute are indeed `Configuration` level. But really the user is going to be manipulating the inner properties of the devices, which can be both `Acquisition` or `Configuration`. These are not shown in the current property pane, but the point kinda still stands.

I think a way around this to define another category defintion

    public abstract class DeviceFactory : Sink<ContextTask>
    {
        internal const string ConfigurationCategory = "Configuration";
        internal const string AcquisitionCategory = "Acquisition";
        internal const string DevicesCategory = "Devices";

        internal abstract IEnumerable<IDeviceConfiguration> GetDevices();
    }

which can be used for member devices of `MultiDeviceFactory`s

image

The `Devices` name is up for suggestions.

bparks13:
I am in favor of adding `Devices` as a category, I was just looking at the docs to see how we refer to these; they are classified as `Device Configuration Operators`, and we refer to them as devices from the aggregate level.

My only caveat with adding this category is that @cjsha and I were discussing what to name the headings in the aggregate page (for instance, [`ConfigureBreakoutBoard`](https://open-ephys.github.io/onix1-bonsai-docs/api/OpenEphys.Onix1.ConfigureBreakoutBoard.html)). Currently they are listed as `Properties`, but if we add this new category I would argue that we change the heading to match it, since the subheadings are the devices.

jonnew:
@bparks13 I agree with all. I feel a small force against this because in the ONI standard, `device` has very particular meaning. However, the ONI standard does not get to uniformly define what a device is for all its possible uses. In the context of this library, its very useful to refer to these leaf objects as devices, so let's do that.

OpenEphys.Onix1/Headstage64OpticalStimulatorTrigger.cs
```diff
@@ -167,6 +177,7 @@ public double InterBurstInterval
         [Editor(DesignTypes.NumericUpDownEditor, DesignTypes.UITypeEditor)]
         [Range(1, int.MaxValue)]
         [Precision(0, 1)]
+        [Category(DeviceFactory.AcquisitionCategory)]
         public uint BurstsPerTrain
         {
             get => burstsPerTrain.Value;
```
jonnew, line 183: (Cant put comment where i want to)
`BurstCurrent` is classified as "Misc". Although it can change in real time, I think i like the fact that is not in the acquisition category since it is readonly.

jonnew (approved):

OpenEphys.Onix1/Headstage64OpticalStimulatorTrigger.cs
bparks13: Agreed, I actually almost put it as `Acquisition`, but for the same reason of it being readonly i left it as Misc

PR #230, opened 2024-08-20
Improve neuropixels calibration file error messages

### LINKED ISSUE #138: ConfigureNeuropixelsV1e node error messages improvements
image
Empty filename path error should be “Gain calibration file must be specified.” or “Adc calibration file must be specified.” to match NPV2e.

image
Wrong calibration file popup does not report probe serial number. Should say “Probe serial number xxx does not match xx calibration file serial number xxx.” to match NPV2e and help the user identify the mistake.

image
Cryptic error message when no DeviceName specified in NeuropixelsV1eData node. Suggestion: “DeviceName must be specified”.

jonnew:
- Remove dangling ")" from NP2.0 missing calibration file message
- Add SN information to NP1.0 exception messages
- Fixes #138

### DIFF
--- a/OpenEphys.Onix1/NeuropixelsV1eRegisterContext.cs
+++ b/OpenEphys.Onix1/NeuropixelsV1eRegisterContext.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using System.IO;
 using System.Linq;
 
 namespace OpenEphys.Onix1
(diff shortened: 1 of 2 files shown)


bparks13 (commented):
These are good changes, however I mistakenly updated these same error messages in #231 while also changing/adding how we parse the file. Currently there will be merge conflicts between the two of them; I suggest we either pick and choose what to keep from #231 and port it here to keep the two PRs completely separate, or we merge these changes with what I did there and close both issues from that PR.

jonnew:
I think merge conflict is fine, we just need to resolve. The conflict will be in the XML tag only, right?

bparks13:
@jonnew There will be merge conflicts outside of the XML tags, I modified the methods so that all parsing for a single calibration file was done inside of a single method, thus allowing me to reuse the same functionality in the GUI. However, we have since decided to remove this file parsing in the GUI until we can check against a schema. I would like to keep those parsing methods intact, as it will make it easier to integrate into the GUI later on, but that conflicts with these changes. It shouldn't be a huge conflict, so I guess it's not a big problem, but I know merge conflicts can be annoying.

bparks13 (approved):
Approving these changes, we can resolve any conflicts as they come up.

jonnew:
@bparks13 I guess I would like to review the changes to the parsing code before merging. Your changes are in the commit history so they can be recalled if needed, espeically if you create and issue with a link the the commit.

PR #236, opened 2024-08-21
Bump version to 0.2.0

bparks13:
To enable the pre-release builds to be correctly installed ahead of released NuGet packages, bump the minor version.


jonnew (approved):
We confirmed that both main and design packages nuget versions were bumped.

PR #239, opened 2024-08-21
Remove link to internal PortController class

bparks13:
When building the docfx website, we do not render private or internal classes, leading to a broken reference.

A broken reference blocks that repo from merging pull requests, therefore we need to remove this reference.

### DIFF
--- a/OpenEphys.Onix1/PortStatus.cs
+++ b/OpenEphys.Onix1/PortStatus.cs
@@ -12,7 +12,7 @@ namespace OpenEphys.Onix1
     /// <remarks>
     /// This data stream class must be linked to an appropriate headstage,
     /// miniscope, etc. configuration whose communication is dictated by
-    /// a <see cref="PortController"/>.
+    /// a PortController.
     /// </remarks>
     [Description("Produces a sequence of port status information.")]
     public class PortStatus : Source<PortStatusFrame>


cjsha:
Fun, my first review in the onix-bonsai-onix1 repo.

I would say, if that's our only mention of PortController in the docs, and we're not going to expand on it any further, and there's no reference where they can learn what it is, I would remove that entire clause from the sentence. 

```
    /// <remarks>
    /// This data stream class must be linked to an appropriate headstage,
    /// miniscope, etc. configuration whose communication is dictated by
    /// a PortController.
    /// </remarks>
```
would turn into something like:
```
    /// <remarks>
    /// This data stream class must be linked to an appropriate headstage,
    /// miniscope, etc. configuration.
    /// </remarks>
```

cjsha (commented):
My feedback exists in my previous comment, but I'm submitting this comment as well to submit my review.

cjsha (approved):
I'm submitting another review with PR approval this time so that whatever you decide to do you can just go ahead and merge it

bparks13:
I agree with removing the whole clause actually, I thought that the language for PortController was listed in other places (i.e., other properties such as `ConfigureHeadstage64PortController`) but they are all internal classes/properties and as such are never served to the website.

PR #243, opened 2024-08-21
Modify NeuropixelsV2e dialogs to scale correctly

### LINKED ISSUE #241: Existing NeuropixelsV2e GUIs do not resize properly at scale values different from 100%
The choice of panels and anchors in the GUI, combined with the development at a scale of 100%, cause controls to shift drastically from where they are supposed to be at scale values of >100%. This can make it seem as if the GUI is nonfunctional, with controls hidden and basically inaccessible.

Switch from using splitter panels to `FlowLayoutPanel` and `TableLayoutPanel`, which can handle DPI and scaling more correctly.

bparks13:
At scale values different from 100% on Windows 11 machines, the `NeuropixelsV2e` dialogs would not scale correctly, leading to hidden controls and making the GUI effectively useless. 

In this PR, all `NeuropixelsV2e` dialogs were updated to use `FlowLayoutPanel` and `TableLayoutPanel` instead of `SplitterPanel`, as the two former panels handle resizing at different scale values more robustly than the latter. 

One consequence of switching to these panels is that they are no longer movable from the user perspective; that is, all panel sizes are locked to their relative size. To view the full filepath for a configuration file, sometimes when maximizing the window it will show the entire path; if that is not enough, the filepath is inside of a textbox that can be selected and stepped through using the arrow keys to view the path.

Fixes #241

### DIFF
--- a/OpenEphys.Onix1.Design/NeuropixelsV2eDialog.Designer.cs
+++ b/OpenEphys.Onix1.Design/NeuropixelsV2eDialog.Designer.cs
@@ -31,17 +31,14 @@
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(NeuropixelsV2eDialog));
             this.menuStrip = new System.Windows.Forms.MenuStrip();
             this.fileToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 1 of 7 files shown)


jonnew:
Yeah upsides far outweigh the downsides. I tested on my laptop oat various zooms without issue. Nice.

jonnew (approved):

PR #256, opened 2024-08-26
Restrict possible driver values

### LINKED ISSUE #178: Consider specifying a list of drivers that users choose from a drop-down
In the `ContextTask`, currently users must type in the driver they wish to use (see [here](https://github.com/neurogears/onix-refactor/blob/4088fd6687c775fdf8534e8aa7de7cb659fbef44/OpenEphys.Onix/OpenEphys.Onix/ContextTask.cs#L55)). To minimize the possibility of mistyping a driver name, it is possible to use [TypeConverter.GetStandardValues](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter.getstandardvalues?view=net-8.0) to define a list of strings that are valid driver names.

To further minimize the possibility of error, overriding the `GetStandardValuesExclusive` method can disable the ability to type in the field, meaning that users are confined to the given list of drivers.

bparks13:
In the `CreateContext` node users could previously type in a string to specify a specific device driver. However, in this PR we restrict the possible values to a drop-down menu that consists of exactly one item: `riffa`.

Fixes #178

### DIFF
--- a/OpenEphys.Onix1/ContextDriverConverter.cs
+++ b/OpenEphys.Onix1/ContextDriverConverter.cs
@@ -1,0 +1,25 @@
+﻿using System.ComponentModel;
+
+namespace OpenEphys.Onix1
+{
+    internal class ContextDriverConverter : StringConverter
+    {
+        public override bool GetStandardValuesSupported(ITypeDescriptorContext context)
+        {
+            return true;
+        }
+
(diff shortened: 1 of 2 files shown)


jonnew (approved):
Tested. Works well.

PR #261, opened 2024-08-27
Remove "a class that" from XML comments

### LINKED ISSUE #209: XML summary sentence structure for classes
I wonder if it's less clear to read something like "This class generates..." or "A class that generates..." ([for example](https://open-ephys.github.io/onix1-bonsai-docs/api/OpenEphys.Onix1.HarpSyncInputData.html)) when it could just be "Generates...". Specifically, they don't need to know the underlying code that comprises the operator is a class. Again, it's this awkward space of straddling API docs and user docs. In this case, I think it's ok 

image

This pattern is repeated throughout

jonnew:
- Unnessesary jargon
- Other minore XML fixes while addressing this issue
- Fixes #209


OpenEphys.Onix1/PortStatus.cs
```diff
@@ -7,10 +7,10 @@
 namespace OpenEphys.Onix1
 {
     /// <summary>
-    /// A class that produces a sequence of port status information frames.
+    /// Produces a sequence of port status information frames.
```
cjsha, line 10: "Produces" => "Generates"? for consistency with the rest and also the name of the method is actually generate

OpenEphys.Onix1/Bno055Data.cs
```diff
@@ -7,13 +7,13 @@
 namespace OpenEphys.Onix1
 {
     /// <summary>
-    /// A class that generates a sequence of 3D orientation measurements produced by BNO055 9-axis inertial measurement unit.
+    /// Generates a sequence of 3D orientation measurements produced by Bno055 9-axis inertial measurement unit.
```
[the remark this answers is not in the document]
cjsha: same with Rhd

OpenEphys.Onix1/BreakoutDigitalInput.cs
```diff
@@ -7,11 +7,11 @@
 namespace OpenEphys.Onix1
 {
     /// <summary>
-    /// A class that produces a sequence of digital input frames from an ONIX breakout board.
+    /// Produces a sequence of digital events from an ONIX breakout board.
```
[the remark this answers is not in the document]
cjsha: I see this is a docs-wide pattern

OpenEphys.Onix1/PortStatus.cs
cjsha: Actually more comments uses the word "produces" instead of "generates". Hm

cjsha (approved):
This PR is approved but please go through and consider all the comments - I'm not sure it requires another review cycle after you address them (though you can re-request me if you want)

OpenEphys.Onix1/PortStatus.cs
jonnew: I did generate for the Generate() method description and Produces for the class description

PR #266, opened 2024-08-27
Add NeuropixelsV2eBeta GUI

### LINKED ISSUE #202: ProbeConfigurationA & ProbeConfigurationB properties lack documentation
I assume that documentation for these NeuropixelsV2eBeta and NeuropixelsV2e properties  are postpone until the configuration GUIs are available for the 0.2.0 release. Either way, I want to note that info is missing in the interim, at least as a reminder.

image

If the probe configuration GUI is sufficiently intuitive, it might suffice to put some instructions on how to use the GUIs in the remarks which get populated here in this table? Not sure how much instruction will be necessary here.

### LINKED ISSUE #227: Add NeuropixelsV2eBeta GUI
Add a GUI that is similar to the `NeuropixelsV2e` GUI.

bparks13:
This PR adds GUIs for the `NeuropixelsV2eBeta` headstage. The added dialogs are functionally similar to the `NeuropixelsV2e` GUIs, but use the correct underlying type (i.e., `*Beta` types). 

Closes #227 
Fixes #202


jonnew (changes requested):
As per our discussion in the meeting today, the feeling is that there must be a way to reuse the non-beta NP2.0 GUI elements instead of all this code reproduction. Perhaps by
- Defining an interface and operating on the interface
- Defining  a common, intermediate base type for both headstages in the main library
- Performing a hard cast instead of testing for types (probably not great)

bparks13:
@jonnew If I try to cast directly using `var configureHeadstage = (ConfigureNeuropixelsV2eHeadstage)component;`, where `component` is of type `ConfigureNeuropixelsV2eBetaHeadstage`, I get an exception.

image

I'll work on implementing a common interface that both headstages implement, then I can pass the interface to the existing dialog instead of copying all of this code.

For cleanliness, I'm going to convert this PR to a draft, and then close it once these new changes are implemented.

bparks13:
This has been implemented much more efficiently in #269. This PR is no longer needed.

PR #269, opened 2024-08-28
Add NeuropixelsV2eBeta GUIs

### LINKED ISSUE #202: ProbeConfigurationA & ProbeConfigurationB properties lack documentation

### LINKED ISSUE #227: Add NeuropixelsV2eBeta GUI

### LINKED ISSUE #268: Create common interface for NeuropixelsV2e and NeuropixelsV2eBeta
While the underlying implementation for both classes requires distinct methods, there are a number of shared members between these two classes that can be implemented in an interface. This has an added benefit of being able to build the `NeuropixelsV2e` dialogs using this interface, and then the `NeuropixelsV2eBeta` types can be passed to the same dialogs to eliminate the need for copy/paste as in #266. 

Once this is implemented, adding functionality for #227 will be quite simple.

bparks13:
This PR adds GUIs for the `NeuropixelsV2eBeta` headstage. The added dialogs call the `NeuropixelsV2e` GUIs, taking advantage of the common interface between `ConfigureNeuropixelsV2e` and `ConfigureNeuropixelsV2eBeta`. When the underlying type is a `*Beta`, all relevant text strings are updated to reflect this.

Closes #227
Closes #268 
Fixes #202

### DIFF
--- a/OpenEphys.Onix1/IConfigureNeuropixelsV2.cs
+++ b/OpenEphys.Onix1/IConfigureNeuropixelsV2.cs
@@ -1,0 +1,11 @@
+﻿namespace OpenEphys.Onix1
+{
+    public interface IConfigureNeuropixelsV2
+    {
+        public bool Enable { get; set; }
+        public NeuropixelsV2QuadShankProbeConfiguration ProbeConfigurationA { get; set; }
+        public string GainCalibrationFileA { get; set; }
+        public NeuropixelsV2QuadShankProbeConfiguration ProbeConfigurationB { get; set; }
+        public string GainCalibrationFileB { get; set; }
+    }
+}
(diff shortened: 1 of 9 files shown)


OpenEphys.Onix1/IConfigureNeuropixelsV2.cs
jonnew, line 1: Please add XML comments to public properties, types, etc. 

Or maybe make them non public?

bparks13: Right, I did forget to add those. I tried to make the interface internal, but that meant it could not be accessed in the Design library.

jonnew (approved):
Ran GUIs, configured channels etc. All seems good.

PR #272, opened 2024-08-29
Fix loading channel configuration files

### LINKED ISSUE #252: Handle exceptions when loading a JSON file
For `NeuropixelsV2e` GUIs, if the JSON file that is being loaded (File -> Load Channel Configuration) is badly formatted or missing necessary fields, the whole dialog crashes. The resulting exception does point the user to a JSON error, hopefully pointing them in the right direction, but I need to handle this during loading properly so the whole dialog doesn't crash.

I have not tested this in `NeuropixelsV1e`, but I am pretty sure this is coming from the common framework so it will affect both dialogs (and any future ones). Hopefully this means that it only needs to be fixed once.

### LINKED ISSUE #271: Channel configurations do not respect file properties on load
`NeuropixelsV2QuadShankProbeConfiguration` and `NeuropixelsV1eProbeConfiguration` both have an issue when loading an existing file. What happens is that the `DeviceChannelIndices` field is pulled out and then saved, but the rest of the fields are discarded. This is due to how I am creating a deep-copy; to fix it, I will need to update `OpenEphys.ProbeInterface.NET` to create clones of the fields, or use the newly modified public `Probe` copy constructor as a way to copy the whole probe without keeping the pointer reference.

This is currently being fixed in the issue-252 branch.

bparks13:
In this PR, channel configuration files that are malformed or are missing required fields no longer throw an exception that forcibly closes the dialog. Exceptions are routed appropriately to notify the user instead. 

Fixes #252

Additionally, files that contain changes other than the device channel indices are now respected. Previously, only the device channel indices were being saved, which meant that channels could be enabled and disabled by reading in a configuration file, but if the probe contour or contact shapes were modified those changes were not reflected in the GUI.

Fixes #271 

Finally, an emergent property of fixing #271 led to a bug in drawing the scale bar after loading files without a probe contour. This has been addressed as well in this PR, but no issue was created for it.

### DIFF
--- a/OpenEphys.Onix1.Design/ChannelConfigurationDialog.cs
+++ b/OpenEphys.Onix1.Design/ChannelConfigurationDialog.cs
@@ -169,10 +169,10 @@ namespace OpenEphys.Onix1.Design
 
             public ProbeEdge(ZedGraphControl zedGraphControl)
             {
-                Left = GetProbeContourMinX(zedGraphControl.GraphPane.GraphObjList);
-                Right = GetProbeContourMaxX(zedGraphControl.GraphPane.GraphObjList);
-                Bottom = GetProbeContourMinY(zedGraphControl.GraphPane.GraphObjList);
-                Top = GetProbeContourMaxY(zedGraphControl.GraphPane.GraphObjList);
+                Left = GetProbeMinX(zedGraphControl.GraphPane.GraphObjList);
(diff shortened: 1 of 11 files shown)


jonnew (approved):
I committed some changes. Please have a look at those to make sure they look ok. After that feel free to merge.

jonnew:
Issue documented in my last commit is opened in #283

PR #278, opened 2024-09-03
Improve StartAqcuisition documentation

### LINKED ISSUE #263: Improve StartAcqusition explaination
This issue stems from this one: https://github.com/open-ephys/onix1-bonsai-docs/issues/19

Basically, we don't have access to the clroni API (and really should not rely on it anyway). We need to just explain in more detail what this operators does.

jonnew:
- Add details to class description about what the operator does and what it produces without cref links to clroni
- Fix typo in BitHelper
- Fixes #263

### DIFF
--- a/OpenEphys.Onix1/StartAcquisition.cs
+++ b/OpenEphys.Onix1/StartAcquisition.cs
@@ -9,19 +9,49 @@ namespace OpenEphys.Onix1
     /// <summary>
     /// Starts data acquisition and frame distribution on a <see cref="ContextTask"/>.
     /// </summary>
+    /// <remarks>
+    /// The <see href="https://open-ephys.github.io/ONI/">open neuro interface (ONI)</see> hardware
+    /// specification and API describe a general purpose acquisition system architecture and programming
+    /// interface for communication with a host PC. One requirement of ONI is a sequence of events that must
+    /// occur in order to start synchronized data acquisition. <see cref="StartAcquisition"/> performs these
+    /// required actions on one or more <see cref="ContextTask"/>s provided in its input sequence. Once
+    /// aquisition is started, devices within a particular context will start to produce data in a format
+    /// called an "ONI frame". The output sequence of this operator is therefore a <see
+    /// cref="GroupedObservable{TKey, TElement}"/>, where
+    /// <list type="table">
+    /// <item>
+    /// <term>Tkey</term>
+    /// <description>
+    /// Is the address of a particular hardware device within a single <see cref="ContextTask"/>.
+    /// </description>
+    /// </item>
+    /// <item>
+    /// <term>TElement</term>
+    /// <description>
+    /// Is a ONI frame produced by the device with address Tkey.
+    /// </description>
+    /// </item>
+    /// </list>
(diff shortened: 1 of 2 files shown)


OpenEphys.Onix1/StartAcquisition.cs
```diff
@@ -31,26 +61,27 @@ public class StartAcquisition : Combinator<ContextTask, IGroupedObservable<uint,
...
-        /// <see cref="oni.Context.Write{T}(uint, T)"/>, and <see cref="oni.Context.Write{T}(uint, T[])"/>. A larger size will reduce
-        /// the average amount of dynamic memory allocation system calls but increase the cost of each of those calls. The minimum
-        /// size of this option is determined by <see cref="ContextTask.MaxWriteFrameSize"/>. The effect on real-timer performance is not as
-        /// large as that of <see cref="ContextTask.BlockReadSize"/>.
+        /// This value determines the amount of memory pre-allocated for calls to <see
+        /// cref="oni.Context.Write(uint, IntPtr, int)"/>, <see cref="oni.Context.Write{T}(uint, T)"/>, and
+        /// <see cref="oni.Context.Write{T}(uint, T[])"/>. A larger size will reduce the average amount of
+        /// dynamic memory allocation system calls but increase the cost of each of those calls. The minimum
+        /// size of this option is determined by <see cref="ContextTask.MaxWriteFrameSize"/>. The effect on
+        /// real-timer performance is not as large as that of <see cref="ContextTask.BlockReadSize"/>.
```
cjsha, line 69: "real-timer" => "real-time".

Also, add more explanation here:

>The effect on real-time performance is not as large as that of \<see cref="ContextTask.BlockReadSize"/\>.

For example:

>This balance is not as critical for real-time performance as that of \<see cref="ContextTask.BlockReadSize"/\> because there tends to be less writes than reads such that the write doesn't consume as much bandwidth. 

or something like that? I mostly want to confirm my understanding of the system, but I think it might also be helpful for the user to understand the different considerations when setting `ReadSize` and `WriteSize`.

OpenEphys.Onix1/StartAcquisition.cs
```diff
@@ -9,19 +9,49 @@ namespace OpenEphys.Onix1
...
+    /// Is the address of a particular hardware device within a single <see cref="ContextTask"/>.
+    /// </description>
+    /// </item>
+    /// <item>
+    /// <term>TElement</term>
+    /// <description>
+    /// Is a ONI frame produced by the device with address Tkey.
+    /// </description>
+    /// </item>
+    /// </list>
```
cjsha, line 34: here's a screenshot so you can see how your xml table is rendered btw:
image
Just to confirm because this might change whether you want to write out the table over a couple sentences instead.
We can also edit its css. If you have any strong opinions about this, we can make issue of it in the docs repo

cjsha (approved):
This PR is approved but please go through and consider all the comments - I'm not sure it requires another review cycle after you address them (though you can re-request me if you want)

OpenEphys.Onix1/StartAcquisition.cs
jonnew: I think its fine. It gets the job done at least.

PR #287, opened 2024-09-04
Improve the shutdown behavior

jonnew:
- This commit addresses two issues
- The first is that if there was an exception setting block read or write sizes, contextConfiguration within ContextTask was not disposed. This lead to a deadlock that required the process to be restarted even if the offending parameter was changed.
- When addressing this, I noticed that there may be a more general issue that is documented on line 314 of ContextTask. The general issue is that contextConfiguration must always be disposed, and there are (unlikely) ways that it might not be.
- Additionally, I improved the error messages presented when a ReadSize or WriteSize exception occurred indicating what value the user needs to use.
- The second issue was that when a hub was configured as a passthrough device, it was not reset to a stanard hub when acqusition was stopped. This could cause issues because the raw device used by passthroughs is a member of hub zero and could lead to huge required block read sizes even a headstage was no longer present. To address this, I simply returned an active disposable whose action is to reset the port to standard mode fro source.CoonfigureHost in ConfigurePortController, rather than Disposable.Empty, which did nothing.


bparks13 (approved):
Looks good, I'm assuming this has been tested and is working appropriately for both the Passthrough State and handling exceptions while creating the context.

Issue #283, opened 2024-09-03
StreamReader arguments in NeuropixelsV1Helper.ParseAdcCalibrationFile() and ParseGainCalibrationFile() should be a path


jonnew:
The `file` input argument in the following functions:  

https://github.com/open-ephys/onix-bonsai-onix1/blob/3fdcec67f5f5d548ddf9125d7859565b55a48d08/OpenEphys.Onix1/NeuropixelsV1Helper.cs#L21

https://github.com/open-ephys/onix-bonsai-onix1/blob/3fdcec67f5f5d548ddf9125d7859565b55a48d08/OpenEphys.Onix1/NeuropixelsV1Helper.cs#L65

should either be a `FileStream` or a `string` indicating a path. StreamReader is to general because calibration data is always provided in a file and this function call expects to start from the beginning of the stream. This will require a change in the public API.

bparks13:
These methods have been marked as `[Obsolete]` in #291, but will not be modified/removed until 1.0.0.

jonnew:
This will be addressed in #304. Im closing since in reality its already addressed in #291.

PR #291, opened 2024-09-05
Add validation of Neuropixels calibration files

### LINKED ISSUE #228: Check that Neuropixels Calibration files match an expected format
Anytime we are opening a file to read data, we should ensure that the file matches an expected format so that it doesn't throw any weird exceptions.

For example, when checking [`NeuropixelsV2e` gain correction files](https://github.com/open-ephys/onix-bonsai-onix1/blob/048e90edb498b3827d97d2e4c6d5e43e4c12cd59/OpenEphys.Onix1/NeuropixelsV2.cs#L116), an incorrectly formatted file leads to weird exceptions that are not very useful for the end user.

There are other areas where we pull data from files without checking if they match some expected format; we can use this issue as a meta-issue to track individual areas where this might be necessary to add some logic for checking file formats.

Related: All GUI functions that are touching the same files should not reimplement logic, but should call these new checking functions and loading functions from the base library.

### LINKED ISSUE #232: Neuropixels V1e and V2e file parsing in the Design library
For `NeuropixelsV1e` and `NeuropixelsV2e`, there are several files that need to be parsed when running the workflow for proper data acquisition.

In the GUIs, we would like to parse these files and display relevant information to the user graphically, but currently the parsing functions can throw exceptions that shut down the GUI immediately. To fix this, we need to create some `NeuropixelsVXCalibrationFile.Validate` methods that take in the filepath, and return `true` or `false` without throwing an exception. If the files pass these validation tests, we can pass to the parsers without fear of an exception being thrown.

From @jonnew: "basically we want to define rigid schemas for each type of file. These can be `nothrow` style functions that just catch exceptions and return true or false if a file can be parsed correctly."

bparks13:
In this PR, the calibration files for `NeuropixelsV1e`, `NeuropixelsV2e`, and `NeuropixelsV2eBeta` are now validated using a custom `TryParse` method. This applies to the ADC / gain calibration files for `NeuropixelsV1e`, and the gain calibration file for `NeuropixelsV2e`/`NeuropixelsV2eBeta`. 

The `TryParse` method will validate that the file exists and can be opened, that there are the correct number of elements, and that each element is able to be parsed into the correct type. For example, the first line in each file should contain a single element, and it should be `ulong` for the serial number. Other elements include the channel number, gain correction value at different gain levels, or ADC values.

Fixes #228 

Related to #283, existing public methods for `NeuropixelsV1e` have been marked as obsolete, since they took a `StreamReader` as an argument instead of the filepath. These methods are not removed or modified, since doing so would be a major revision.

In addition to creating the `TryParse` method, this PR also takes advantage of these new methods by adding the correction values back into the various dialogs so that the user can view the ADC values and gain corrections that will be applied to the data based on the current gains selected. These are added at all dialog levels, so that anytime a file is selected, the gain will be updated in a textbox. For ADC values, instead of being in a tab like they were previously, they are now opened in a new window when a button is pressed. This window is modal, so the rest of the GUI is nonfunctional until the ADC window is closed. 

Fixes #232 

While adding control elements, including a status bar to display the serial number, there were some small quality-of-life changes made to control sizes.

### DIFF
--- a/OpenEphys.Onix1/NeuropixelsV1Helper.cs
+++ b/OpenEphys.Onix1/NeuropixelsV1Helper.cs
@@ -9,7 +9,164 @@ namespace OpenEphys.Onix1
     /// </summary>
     public static class NeuropixelsV1Helper
     {
-        internal const int NumberOfGains = 8;
+        internal const int NumberOfGainFactors = 8;
+        internal const int NumberOfAdcParameters = 8;
+
+        /// <summary>
+        /// Tries to parse the ADC calibration file.
+        /// </summary>
+        /// <remarks>
+        /// Checks if the given filename points to a real file, and checks each individual element:
+        /// <para>
+        /// 1) Serial number as the first line.
+        /// </para>
+        /// <para>
+        /// 2) All remaining lines (32 lines) contain the ADC correction values for each ADC. First is the ADC number,
+        /// followed by the <see cref="NeuropixelsV1eAdc"/> values. All elements are separated by commas, and each element is a valid 
+        /// integer value.
+        /// </para>
+        /// </remarks>
+        /// <param name="adcCalibrationFile">String defining the path to the ADC calibration file.</param>
+        /// <returns><see cref="NeuropixelsV1eAdcCalibration"/> object that contains the ADC calibration values. This object is null if the file was not successfully parsed.</returns>
+        public static NeuropixelsV1eAdcCalibration? TryParseAdcCalibrationFile(string adcCalibrationFile)
+        {
+            if (!File.Exists(adcCalibrationFile)) return null;
+
+            var lines = File.ReadLines(adcCalibrationFile);
+
+            if (lines.Count() != NeuropixelsV1e.AdcCount + 1) return null;
+            if (!ulong.TryParse(lines.ElementAt(0), out var serialNumber)) return null;
+
+            if (!lines
+                .Skip(1)
+                .Select(l =>
+                {
+                    var ok = int.TryParse(l.Split(',')[0], out int adcNumber);
+                    return (Ok: ok, AdcNumber: adcNumber);
+                })
+                .Where(l => l.Ok)
+                .Select(l => l.AdcNumber)
+                .SequenceEqual(Enumerable.Range(0, NeuropixelsV1e.AdcCount))) return null;
+
+            var adcs = lines
+                       .Skip(1)
+                       .Select(l =>
+                       {
+                           var calibrationValues = l
+                                                   .Split(',')
+                                                   .Skip(1)
+                                                   .Select(p =>
+                                                   {
+                                                       var ok = int.TryParse(p, out int param);
+                                                       return (Ok: ok, Param: param);
+                                                   })
+                                                   .Where(l => l.Ok)
+                                                   .Select(l => l.Param);
+
+                           return calibrationValues.Count() == NumberOfAdcParameters
+                           ? null
+                           : new NeuropixelsV1eAdc {
+                               CompP = calibrationValues.ElementAt(0),
+                               CompN = calibrationValues.ElementAt(1),
+                               Slope = calibrationValues.ElementAt(2),
+                               Coarse = calibrationValues.ElementAt(3),
+                               Fine = calibrationValues.ElementAt(4),
+                               Cfix = calibrationValues.ElementAt(5),
+                               Offset = calibrationValues.ElementAt(6),
+                               Threshold = calibrationValues.ElementAt(7)
+                           };
+                       });
+
+            return adcs.Any(adc => adc == null) ? null : new(serialNumber, adcs.ToArray());
+        }
+
+        /// <summary>
+        /// Tries to parse the gain calibration file.
+        /// </summary>
+        /// <remarks>
+        /// Checks if the given filename points to a real file, and checks each individual element:
+        /// <para>
+        /// 1) Serial number as the first line.
+        /// </para>
+        /// <para>
+        /// 2) All remaining lines (384 lines) contain the gain correction values for each channel. First is the channel number,
+        /// followed by the <see cref="NeuropixelsV1Gain"/> values. All elements are separated by commas, and each element is a valid 
+        /// double value. After the channel number, the first 8 values are AP gain corrections, related to the spike-band, and the last 8 values
+        /// are LFP gain corrections, for the LFP band.
+        /// </para>
+        /// </remarks>
+        /// <param name="gainCalibrationFile">String defining the path to the gain calibration file.</param>
+        /// <param name="apGain">Current <see cref="NeuropixelsV1Gain"/> for the AP data.</param>
+        /// <param name="lfpGain">Current <see cref="NeuropixelsV1Gain"/> for the LFP data.</param>
+        /// <returns><see cref="NeuropixelsV1eGainCorrection"/> object that contains the AP and LFP gain correction values. This object is null if the file was not successfully parsed.</returns>
+        public static NeuropixelsV1eGainCorrection? TryParseGainCalibrationFile(string gainCalibrationFile, NeuropixelsV1Gain apGain, NeuropixelsV1Gain lfpGain)
+        {
+            if (!File.Exists(gainCalibrationFile)) return null;
+
+            var lines = File.ReadLines(gainCalibrationFile);
+
+            if (lines.Count() != NeuropixelsV1e.ElectrodeCount + 1) return null;
+            if (!ulong.TryParse(lines.ElementAt(0), out var serialNumber)) return null;
+
+            if (!lines
+                .Skip(1)
+                .Select(l =>
+                {
+                    var ok = int.TryParse(l.Split(',')[0], out var channel);
+                    return (Ok: ok, Channel: channel);
+                })
+                .Where(l => l.Ok)
+                .Select(l => l.Channel)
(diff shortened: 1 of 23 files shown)


jonnew (changes requested):
I added some suggestions for improving the conciseness and strictness of the code. They pertain to np 2.0 but the same suggestions can brough to 1.0 if you think they are good changes.

OpenEphys.Onix1/NeuropixelsV2Helper.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// 1) Serial number as the first line.
+        /// </para>
+        /// <para>
+        /// 2) All remaining lines (384 lines) are an integer and a double value, separated by a comma.
+        /// The integer defines the channel number, and the double is the gain calibration value for that channel.
+        /// </para>
+        /// </remarks>
+        /// <param name="gainCalibrationFile">String containing the path to the gain calibration file.</param>
+        /// <returns><see cref="NeuropixelsV2GainCorrection"/> object that contains the gain correction value. This object is null if the file was not successfully parsed.</returns>
+        public static NeuropixelsV2GainCorrection? TryParseGainCalibrationFile(string gainCalibrationFile)
```
jonnew, line 25: Apparently I didn't add this comment to my lat review. I'm retyping it now.

This funciton can be made stricter and my consise by using Linq. Here is an example that checks if the file has the right number of lines, that channels in the file are in order and that every gain value is the same, all of which are assumed by us when performing data correction:

```csharp
public static NeuropixelsV2GainCorrection? TryParseGainCalibrationFile(string gainCalibrationFile)
{
    if (!File.Exists(gainCalibrationFile)) return null;
    var lines = File.ReadLines(gainCalibrationFile);
    if (lines.Count() != 385) return null;
    if (!ulong.TryParse(lines.ElementAt(0), out var serialNumber)) return null;

    if (!lines
        .Skip(1)
        .Select(l =>
        {
            var s = l.Split(',')[0];
            int.TryParse(s, out var channel);
            return channel;
        })
        .SequenceEqual(Enumerable.Range(0, 384))) return null;

    var gains = lines.Skip(1)
                     .Select(l =>
                     {
                         var s = l.Split(',')[1];
                         double.TryParse(s, out var gainCorrection);
                         return gainCorrection;
                     })
                     .Distinct();

    return gains.Count() == 1 ? new(serialNumber, gains.ElementAt(0)) : null;
}
```

Similar technique can be brought to 1,0 parsing stuff

OpenEphys.Onix1/NeuropixelsV1Helper.cs
```diff
@@ -10,6 +11,150 @@ namespace OpenEphys.Onix1
...
+                                    Fine = calibrationValues.ElementAt(4),
+                                    Cfix = calibrationValues.ElementAt(5),
+                                    Offset = calibrationValues.ElementAt(6),
+                                    Threshold = calibrationValues.ElementAt(7)
+                                };
+                            });
+
+            if (adcs.Any(adc => adc == null)) return null;
+
+            return new(serialNumber, adcs.ToArray());
```
jonnew, line 84: Why does lazyness (see below) matter with linq? Because linq is lazy. Right here, in `ToArray()` linq will say, "dang it, i cant just hold a bunch iterators anymore I need to actually come up with something." It will ask the for the first element. Lets say that's badly formatted, then the null check above will fail and the file will be closed after reading a single line. 

For applications that don't need determinism and want to reduce their memory footprint, lazy evaluation is superior.

OpenEphys.Onix1/NeuropixelsV1Helper.cs
```diff
@@ -10,6 +11,150 @@ namespace OpenEphys.Onix1
...
+                    int.TryParse(s, out var channel);
+                    return channel;
+                })
+                .SequenceEqual(Enumerable.Range(0, NeuropixelsV1e.ElectrodeCount))) return null;
+
+            var apIndex = Array.IndexOf(Enum.GetValues(typeof(NeuropixelsV1Gain)), apGain) + 1;
+            var apGainCorrections = lines.Skip(1)
+                                         .Select(l =>
+                                         {
+                                             var s = l.Split(',')[apIndex];
```
jonnew, line 134: This will throw if apIndex is out of range

OpenEphys.Onix1/NeuropixelsV1Helper.cs
```diff
@@ -10,6 +11,150 @@ namespace OpenEphys.Onix1
...
+        /// </para>
+        /// <para>
+        /// 2) All remaining lines (32 lines) contain the ADC correction values for each ADC. First is the ADC number,
+        /// followed by the <see cref="NeuropixelsV1eAdc"/> values. All elements are separated by commas, and each element is a valid 
+        /// integer value.
+        /// </para>
+        /// </remarks>
+        /// <param name="adcCalibrationFile">String defining the path to the ADC calibration file.</param>
+        /// <returns><see cref="NeuropixelsV1eAdcCalibration"/> object that contains the ADC calibration values. This object is null if the file was not successfully parsed.</returns>
+        public static NeuropixelsV1eAdcCalibration? TryParseAdcCalibrationFile(string adcCalibrationFile)
```
jonnew, line 32: Here is a replacement function that takes some of what you've done and gets rid of the unnecessary exception catching and magic return codes. 

``` csharp
        public static NeuropixelsV1eAdcCalibration? TryParseAdcCalibrationFile(string adcCalibrationFile)
        {
            if (string.IsNullOrEmpty(adcCalibrationFile)) return null;
            if (!File.Exists(adcCalibrationFile)) return null;

            var lines = File.ReadLines(adcCalibrationFile); 

            if (lines.Count() != NeuropixelsV1e.AdcCount + 1) return null;
            if (!ulong.TryParse(lines.ElementAt(0), out var serialNumber)) return null;

            if (!lines
                .Skip(1)
                .Select(l =>
                {
                    var ok = int.TryParse(l.Split(',')[0], out var adcNumber);
                    return (Ok: ok, Adc: adcNumber);
                })
                .Where(l => l.Ok)
                .Select(l => l.Adc)
                .SequenceEqual(Enumerable.Range(0, NeuropixelsV1e.AdcCount))) return null;

            var adcs = lines.Skip(1)
                            .Select(l =>
                            {
                                var calibrationValues = l.Split(',')
                                                         .Skip(1)
                                                         .Select(e =>
                                                         {
                                                             var ok = int.TryParse(e, out int param);
                                                             return (Ok: ok, Param: param);
                                                         })
                                                         .Where(l => l.Ok)
                                                         .Select(l => l.Param);

                                return calibrationValues.Count() != NumberOfAdcParameters
                                    ? null
                                    : new NeuropixelsV1eAdc {
                                        CompP = calibrationValues.ElementAt(0),
                                        CompN = calibrationValues.ElementAt(1),
                                        Slope = calibrationValues.ElementAt(2),
                                        Coarse = calibrationValues.ElementAt(3),
                                        Fine = calibrationValues.ElementAt(4),
                                        Cfix = calibrationValues.ElementAt(5),
                                        Offset = calibrationValues.ElementAt(6),
                                        Threshold = calibrationValues.ElementAt(7)
                                    };
                            });

            return adcs.Any(adc => adc == null) ?  null : new(serialNumber, adcs.ToArray());
        }
```

OpenEphys.Onix1/NeuropixelsV1Helper.cs
```diff
@@ -10,6 +11,150 @@ namespace OpenEphys.Onix1
...
+                                         .Distinct();
+
+            if (apGainCorrections.Count() == 2) apGainCorrections = apGainCorrections.Where(val => val != 1.0);
+            if (apGainCorrections.Count() != 1) return null;
+
+            var lfpIndex = Array.IndexOf(Enum.GetValues(typeof(NeuropixelsV1Gain)), lfpGain) + 1 + NumberOfGains;
+            var lfpGainCorrections = lines.Skip(1)
+                                          .Select(l =>
+                                          {
+                                              var s = l.Split(',')[lfpIndex];
```
jonnew, line 147: This will throw if apIndex is out of range.

OpenEphys.Onix1/NeuropixelsV1Helper.cs
```diff
@@ -10,6 +11,150 @@ namespace OpenEphys.Onix1
...
+        /// followed by the <see cref="NeuropixelsV1Gain"/> values. All elements are separated by commas, and each element is a valid 
+        /// double value. After the channel number, the first 8 values are AP gain corrections, related to the spike-band, and the last 8 values
+        /// are LFP gain corrections, for the LFP band.
+        /// </para>
+        /// </remarks>
+        /// <param name="gainCalibrationFile">String defining the path to the gain calibration file.</param>
+        /// <param name="apGain">Current <see cref="NeuropixelsV1Gain"/> for the AP data.</param>
+        /// <param name="lfpGain">Current <see cref="NeuropixelsV1Gain"/> for the LFP data.</param>
+        /// <returns><see cref="NeuropixelsV1eGainCorrection"/> object that contains the AP and LFP gain correction values. This object is null if the file was not successfully parsed.</returns>
+        public static NeuropixelsV1eGainCorrection? TryParseGainCalibrationFile(string gainCalibrationFile, NeuropixelsV1Gain apGain, NeuropixelsV1Gain lfpGain)
```
jonnew, line 106: Have a look at adc parser for inspiration on how this can be improved

jonnew (changes requested):

jonnew (approved):
I made some minor modifications. This is ready to merge, I think.

PR #292, opened 2024-09-06
Explicitly link `Trigger` operators to their constituent `Configure` operator

### LINKED ISSUE #211: Headstage64 ElectricalStimulator & OpticalStimulator Remarks require more information
There should be more instructions on how to use the ``ElectricalStimulator`` and ``OpticalStimulator`` properties on the ConfigureHeadstage64. I think it would be proper to include more info in their remarks? in addition to other pages in the docs.

image

Also, I know these properties don't necessarily belong in the Misc subheader, this is its own [issue that belongs to the docs repo](https://github.com/open-ephys/onix1-bonsai-docs/issues/11).

bparks13:
Currently there is no indication of how to link a `Trigger` operator to the `Configure` operator. By adding remarks in this PR, this linking is now explicitly stated.

Fixes #211

### DIFF
--- a/OpenEphys.Onix1/ConfigureHeadstage64.cs
+++ b/OpenEphys.Onix1/ConfigureHeadstage64.cs
@@ -81,6 +81,7 @@ namespace OpenEphys.Onix1
         /// <summary>
         /// Gets or sets onboard electrical stimulator configuration.
         /// </summary>
+        /// <inheritdoc cref="ConfigureHeadstage64ElectricalStimulator"/>
         [Category(DevicesCategory)]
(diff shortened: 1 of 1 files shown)


cjsha (approved):
This doesn't affect the docs immediately because some changes need be made to the template in order to pull in these remarks in and use them effectively. However, I examined the raw data model generated by docfx, and the remarks content generated by the <inheritdoc> tag looks exactly like what the `OpticalStimulator` and `ElectricalStimulator` sections in the `ConfigureHeadstage64` page need.

PR #293, opened 2024-09-06
Fix broken HarpSyncInput reference in xml comment

cjsha:
- HarpSyncInput does not exist; HarpSyncInputData does however


bparks13 (approved):
Small correction; HarpSyncInput does exist, which is why it is a valid XML comment and is not flagged by the IDE. However, it is an internal class so it is not exposed by the website. Either way, this change fixes the link to point to the correct operator.

PR #301, opened 2024-09-16
Remove Breakout prefix from classes where its redundant

### LINKED ISSUE #297: Remove `Breakout` from names of static hardware definition classes within `Configure<>` classes
For example, remove the `Breakout` prefix here

https://github.com/open-ephys/onix-bonsai-onix1/blob/f48d2f2f7a3af84d840c6a5d696bf5cb99d5616c/OpenEphys.Onix1/ConfigureBreakoutAnalogIO.cs#L282C18-L282C34

Why? Because it was wrong to add `Breakout` all over the place in the first place, including to  `ConfigureBreakoutAnalogIO` and `ConfigureBreakoutDigitalIO`. The addition of this breakout grouping information should always have come from inclusion of these configuration types in the `BreakoutBoard` `MutiDeviceFractory`.  This was correct because these devices dont actually even exist on the breakout board and that situation is perectly modeled by the `MultiDeviceFactory` which is a logical grouping of components that dont actually need to correspond to anything physical. Further, when `Breakout` is added to the type name, you end up with stupid looking `DeviceNames` like `BreakoutBoard/BreakoutAnalogIO`.  Doing the full change will break the public API. In the meantime, at least the inner, static static hardware definition classes within `Configure<>` classes can be changed dos that the drop down list items don't look so bad.

### LINKED ISSUE #300: Change `BreakoutAnalogInput` to `AnalogInput`
This same change can be applied to `BreakoutAnalogOutput`, `BreakoutDigitalInput`, and `BreakoutDigitalOutput`. 

@jonnew mentioned this to me off-handedly that he doesn't like the current naming convention and I think it's worth discussing. It seems redundant, and I'm not sure what the benefit is.

jonnew:
- Preserve backwards compatibility with old workflows.
- ~This commit requires a major revision increment~ Because we are in the 0 major revision, this change is permitted, even though its API breaking
- Fixes #297 
- Fixes #300 
- After this is merged, I will create 1.0.0 issue requesting the removal of obsolete types

### DIFF
--- a/OpenEphys.Onix1/ConfigureDigitalIO.cs
+++ b/OpenEphys.Onix1/ConfigureDigitalIO.cs
@@ -3,7 +3,8 @@ using System.ComponentModel;
 
 namespace OpenEphys.Onix1
 {
-    [Obsolete]
+    /// <inheritdoc cref = "ConfigureDigitalIO"/>
+    [Obsolete("Use ConfigureDigitalIO instead. This operator will be removed in version 1.0.0v")]
     public class ConfigureBreakoutDigitalIO : ConfigureDigitalIO { }
 
     /// <summary>
--- a/OpenEphys.Onix1/ConfigureDigitalIO.cs
+++ b/OpenEphys.Onix1/ConfigureDigitalIO.cs
@@ -3,22 +3,25 @@ using System.ComponentModel;
 
 namespace OpenEphys.Onix1
 {
+    [Obsolete]
+    public class ConfigureBreakoutDigitalIO : ConfigureDigitalIO { }
+
     /// <summary>
     /// Configures the ONIX breakout board's digital inputs and outputs.
     /// </summary>
     /// <remarks>
     /// This configuration operator can be linked to data IO operators, such as <see
-    /// cref="BreakoutDigitalInput"/> and <see cref="BreakoutDigitalOutput"/>, using a shared
+    /// cref="DigitalInput"/> and <see cref="DigitalOutput"/>, using a shared
     /// <c>DeviceName</c>.
     /// </remarks>
     [Description("Configures the ONIX breakout board's digital inputs and outputs.")]
-    public class ConfigureBreakoutDigitalIO : SingleDeviceFactory
+    public class ConfigureDigitalIO : SingleDeviceFactory
     {
         /// <summary>
-        /// Initialize a new instance of the <see cref="ConfigureBreakoutDigitalIO"/> class.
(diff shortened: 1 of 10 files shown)


OpenEphys.Onix1/ConfigureDigitalIO.cs
```diff
@@ -3,22 +3,25 @@
 
 namespace OpenEphys.Onix1
 {
+    [Obsolete]
+    public class ConfigureBreakoutDigitalIO : ConfigureDigitalIO { }
```
bparks13, line 7: Missing XML comment

OpenEphys.Onix1/DigitalInput.cs
```diff
@@ -6,18 +6,21 @@
 
 namespace OpenEphys.Onix1
 {
+    [Obsolete]
+    public class BreakoutDigitalInput : DigitalInput { }
```
bparks13, line 10: Missing XML comment

OpenEphys.Onix1/DigitalOutput.cs
```diff
@@ -6,32 +6,35 @@
 
 namespace OpenEphys.Onix1
 {
+    [Obsolete]
+    public class BreakoutDigitalOutput : DigitalOutput { }
```
bparks13, line 10: Missing XML comment

bparks13 (approved):
I see no issues with this PR. There are some missing XML comments from the newly created obsolete classes, but those are easy to fix. 

Visually, if someone opens up an existing workflow that has one of these obsolete nodes, it shows the correct color but appears with cross-hatches to indicate that something is up. 

image

It still builds correctly, shows the correct output types, and does not appear to throw any errors.

jonnew:
Thanks for the review. The big issue we need to consider is do we put effort into maintaining backwards compatibility when we are in version 0.X.X?  Im in the middle of refactoring the Bno055 to create a common class for all headstages that perform polling. Should I add these obsolete derived classes there as well? I guess so?

bparks13:
I guess the question for BNO is what is the opportunity cost for adding the obsolete derived classes? I'm guessing there aren't many workflows that depend on this yet, but I think it would be good to add the derived classes if it is easy to do, and then remove the obsolete classes once we reach 1.0

jonnew:
That's precisely what I was thinking as well. I will add them to the refactored Bno stuff. I also realized you can put a note in the `[Obsolete]` attribute. We will state there they will be removed in 1.0.0

PR #303, opened 2024-09-17
Combine all polled Bno055 classes into common type

### LINKED ISSUE #281: Consolidate passthrough BNO055 configure and data operators
The BNO055 has a unique configuration and data operator for each of the passthrough-based headstage it works with. For example:

https://github.com/open-ephys/onix-bonsai-onix1/blob/3fdcec67f5f5d548ddf9125d7859565b55a48d08/OpenEphys.Onix1/ConfigureNeuropixelsV2eBno055.cs#L11

There are others in
- Miniscope
- Neuropixels V1
- Neuropixels V2 Beta

This seems very redundant. Why can they not be consolidated? will in be confusing that `ConfigureBno055` and `Bno055Data` is not he same as whatever consolidated class we come up with?

jonnew:
- ConfigurePolledBno055 and PolledBno055Data classes replace headstage specific versions of the same thing. Axis map is defined via a couple properties that are not browsable in editor
- Old configuration and data operators are marked obsolete
- Fixes #281


OpenEphys.Onix1/ConfigurePolledBno055.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        public bool Enable { get; }
+    }
+
+    /// <summary>
+    /// Specifies the axis map of a Bno055 IMU
+    /// </summary>
+    /// <remarks>
+    /// The axis of the device can be reconfigured to the new reference axis to account for
+    /// differences it its mounting position. The following values can be applied to the Bno055's
```
bparks13, line 143: "differences it its" => "differences i**n** its"

OpenEphys.Onix1/ConfigureNeuropixelsV1eHeadstage.cs
```diff
@@ -43,16 +43,17 @@ public ConfigureNeuropixelsV1eHeadstage()
         [Category(DevicesCategory)]
         [TypeConverter(typeof(SingleDeviceFactoryConverter))]
         [Description("Specifies the configuration for the NeuropixelsV1e device.")]
-        public ConfigureNeuropixelsV1e NeuropixelsV1e { get; set; } = new();
+        public ConfigureNeuropixelsV1e NeuropixelsV1e { get; set; } = new();
```
bparks13, line 46: nit: Extra whitespace

OpenEphys.Onix1/ConfigurePolledBno055.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// </remarks>
+        [Category(ConfigurationCategory)]
+        [Description("Specifies whether the Bno055 device is enabled.")]
+        public bool Enable { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets the axis map that will be applied during configuration.
+        /// </summary>
+        [Browsable(false)]
+        public Bno055AxisMap AxisMap { get; set; }
```
bparks13, line 55: Since these are not browsable they can't be modified by the user in Bonsai, but should these be settable? These seem like they should not be changeable once configured.

jonnew: This is completely resolved in latest commit.

PR #315, opened 2024-09-29
Functional onix-headstage-neuropix1 support

jonnew:
- Needs XML docs
- Needs GUI integration
- Needs testing with more than one probe


jonnew:
@bparks13 
- I removed the "e" suffix in a bunch of NeuropixelsV1**e** classes that are actually generic across headstages. 
- This headstage is needs GUI integration
- It also needs xml docs, but I can do that once you've done the GUI work.

jonnew:
It seems like there is a little left over icon at the bottom of the GUI still:

image

jonnew:
When i run the GUI on a ConfigureNeuropixelsV1f (no headstage), there is a "Probe A" indicator in the window title:

image

I dont think that should be there because the A vs. B is given by the device address in this case.

bparks13:
@jonnew The icon is supposed to be there, it indicates if the calibration files are valid, and if the serial numbers of the files match. This does not touch hardware, but is meant as an easy way to see which files are currently loaded and if they match.

However, the text is being cut off, which is incorrect, so that needs to be fixed.

bparks13 (approved):
Looks good, ready for a final review with hardware and testing the GUI.

jonnew:
Looks good now, however, the NP2.0 GUIs show this:

image

it would be nice if it had the same look as NP1.0:

image

jonnew:
@aacuevas

bparks13:
@jonnew the 2.0 dialogs have been updated to match these changes to 1.0

PR #332, opened 2024-10-13
Tweaks for passthrough devices and miniscope fixes

### LINKED ISSUE #324: I2C rate control should be done on the serializer
On the miniscope, rate control is set through i2c calls to the deserializer, when it is the serializer registers which should be set

https://github.com/open-ephys/bonsai-onix1/blob/cd87fc45d58b87e92598ff037144c3f98a5befd2/OpenEphys.Onix1/ConfigureUclaMiniscopeV4Camera.cs#L195

### LINKED ISSUE #325: DS90UB9xSerializerI2CRegister are valid for DS90UB953 only
The `DS90UB9x.DS90UB9xSerializerI2CRegister` enum contains addresses for i2c access to the remote serializer.
There particular addresses are valid for DS90UB953 serializers, but invalid for the 933 or 913 variants, so headstages using these variants are not properly set-up.

There should be to enums, one for 953 and another for 913/933 variants. Since we know which chip each headstage is using, it should be straightforward to just use the appropriate adress enum as needed.

### LINKED ISSUE #326: SERDES link might not be fully initialized on passthrough devices
Some of the critical SERDES startup parameters are initialized by hardware. On the old library, we relied on the implicit order of operations for the hardware to set them up before switching to passthrough mode. In the new library, I suspect it might not always be the case.

It would be generally safe to run the standard initialization sequence on software when initializing a passthrough device, to ensure correct functionality.

### LINKED ISSUE #327: On miniscope, waiting for pll to settle might be causing instability
This is a purely empirical result done on a very limited set of devices.

However, I have noticed that when removing the waiting times on PLL activation, set on:
https://github.com/open-ephys/bonsai-onix1/blob/cd87fc45d58b87e92598ff037144c3f98a5befd2/OpenEphys.Onix1/ConfigureUclaMiniscopeV4Camera.cs#L222

register access to the atmega registers become much more reliable.

The old library did not include these waits, so they might be creating some unintended consequence.

### LINKED ISSUE #328: On miniscope headstage, a small wait before checking link state might help
By the ds90ub9xx serdes specification, LOCK state might take up to a hundred milliseconds to appear after all settings have been set

As such, it would be helpful to add such a wait state before checking the lock register at:
https://github.com/open-ephys/bonsai-onix1/blob/cd87fc45d58b87e92598ff037144c3f98a5befd2/OpenEphys.Onix1/ConfigureUclaMiniscopeV4.cs#L168

For reference, this is a wait state that existed on the old library, I believe it does not hurt to import it to here, specially since this one stems from a documented constraint.

### LINKED ISSUE #329: In miniscope power discovery, Configuredeserializer should be called inside try block
This line
https://github.com/open-ephys/bonsai-onix1/blob/cd87fc45d58b87e92598ff037144c3f98a5befd2/OpenEphys.Onix1/ConfigureUclaMiniscopeV4.cs#L156

should be called inside the following try block. This, however, includes lines that write into the remote serializer. Since this is called during power discovery, it could be that these get called with insuficient power, resulting in a failed read. Thus, either the whole `ConfigureDeserializer` method should be called inside the try block, or the serializer-related lines should be called on a different method, inside the try block

aacuevas:
This aims to solve some issues regarding miniscope initialization, but adds some tweaks that can potentially prevent issues on neuropixel devices as well

- Adds a static method for initializing the serdes link
- Adds a static method for setting serializer i2c rate speed easily
- Changed the serializer register enums for a full name `DS90UB933SerializerRegisters` , and include enum for 953 variant, removing any possible ambiguity since register addresses are different.
- Miniscope I2C rate is set to 80KHz, which is the value that it had in the latest 0.3 release and which worked at SfN. We can experiment with rates to find the optimum one.
- Fixes #324
- Fixes #325 
- Fixes #326 
- Fixes #327 
- Fixes #328
- Fixes #329

I thought about creating a proper "serdes link" managing class which had methods to access GPIOs and all, but I decided against it. Long-term plan is to remove passthrough access altogether, so any extra work, which would have required device logic changes as well, seems unneeded.

### DIFF
--- a/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
+++ b/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
@@ -166,8 +166,8 @@ namespace OpenEphys.Onix1
                 var deviceInfo = new NeuropixelsV1eDeviceInfo(context, DeviceType, deviceAddress, probeControl);
                 var shutdown = Disposable.Create(() =>
                 {
-                    serializer.WriteByte((uint)DS90UB933SerializerI2CRegister.GPIO10, NeuropixelsV1e.DefaultGPO10Config);
-                    serializer.WriteByte((uint)DS90UB933SerializerI2CRegister.GPIO32, NeuropixelsV1e.DefaultGPO32Config);
+                    serializer.WriteByte((uint)DS90UB933SerializerI2CRegister.Gpio10, NeuropixelsV1e.DefaultGPO10Config);
+                    serializer.WriteByte((uint)DS90UB933SerializerI2CRegister.Gpio32, NeuropixelsV1e.DefaultGPO32Config);
                 });
                 return new CompositeDisposable(
                     DeviceManager.RegisterDevice(deviceName, deviceInfo),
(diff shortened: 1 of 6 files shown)


aacuevas:
As a note, this is mostly what was implemented on the SfN demo, with proper cleanups and methods and including the serdes init logic on the neuropixel probes as well.
I would recommend testing in case any typo might have introduced a passtrhough-wide bug, but overall should be the same logic.

jonnew:
Has nice added benefit of really speeding of npx configuration with i2c running at 400 kHz.

jonnew:
Tests with miniscope V4 indicate big increase in reliability. Could not get it to fail.

PR #340, opened 2024-10-18
Reduced I2C clock when talking to atMega on miniscope

### LINKED ISSUE #334: Can miniscope i2c be switched to 400 kHz post configuration?
Miniscope i2c is turned to 80 kHz during configuration to increase reliability. However, this limits BNO055 sample rate. Does returning it to 400 kHz following configuration affect stability?

jonnew:
- Only set reduced (80 kHz) I2C clock when talking to the atMega, which has a bit-banged implementation
- For the rest of the chips, reset to to 200 kHz, which seems to provide max sampling on BNO while maintaining reliability
- Fixes #334

### DIFF
--- a/OpenEphys.Onix1/ConfigureUclaMiniscopeV4.cs
+++ b/OpenEphys.Onix1/ConfigureUclaMiniscopeV4.cs
@@ -22,8 +22,7 @@ namespace OpenEphys.Onix1
     /// </remarks>
     public class ConfigureUclaMiniscopeV4 : MultiDeviceFactory
     {
-
-        const double MaxVoltage = 5.6;
+        const double MaxVoltage = 5.7;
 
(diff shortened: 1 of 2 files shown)


aacuevas (approved):
The only weird interaction that I can see is that, when changing a miniscope parameter (e.g.: led brightness), the threading model could: Set I2C SCLHigh value, poll a BNO value, then set the SCLLow and perform other operations. However, this should not be an issue, it would be akin to clock stretching, which already happens.

It should be documentes, nonetheless, that polledBNO does not guarantee pure periodicity and that it might be affected even further by this sporadic parameter changes.

Apart for this specific thing, which we just need to be aware of, I think this is a good overall solution. I assume it was tested in hardware as well.

PR #342, opened 2024-10-21
Fixed step size logic for Rhs2116 dialogs

bparks13:
Update logic for choosing step sizes. Also fixed how values are read from existing stimuli.

Updated error message to be more user-friendly, focuses on the amplitude (which is user-defined) instead of the step size (which is automatically selected based on the amplitude).

Fixed alignment of `Apply` button.


jonnew (approved):
Review performed during meeting and by playing with GUI afterwards

PR #346, opened 2024-10-22
Rhs2116 Stimulus Sequence Dialog - Plot pulses in red to highlight them

### LINKED ISSUE #343: Rhs2116 Stimulus Sequence Dialog - Plot pulses with a red overlay to highlight them
After plotting the stimulus sequence waveform, plot an overlay where every pulse is highlighted with a red line to emphasize it. This is especially useful for relatively small pulses, which can be hard to visualize when fully zoomed out.

This will probably be easiest to do by building an array or list that contains the XY coordinates of the pulses as they are added to the full waveform, and then plotting them in a separate method. Need to make sure that the Z-order is correct so that the red is on top of the blue lines.

bparks13:
In addition to plotting the whole stimulus waveform, overlay the individual pulses in red to highlight them, making it easier to note the timings of individual pulses.

If pulses are too small (i.e., only has one step for the amplitude), it can be hard to visualize them still.

- Fixes #343

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <PackageIcon>icon.png</PackageIcon>
-    <VersionPrefix>0.4.0</VersionPrefix>
+    <VersionPrefix>0.4.1</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <LangVersion>10.0</LangVersion>
     <Features>strict</Features>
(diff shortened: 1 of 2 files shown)


jonnew:
This may or may not be related to this PR: the performance of plotting the waveforms seems to be poor enough that if a signficant number of stimuli are added, the GUI becomes nearly unresponsive. In the following video im trying to select channels, zoom, etc and there are multiple seconds between actions and refresh:


https://github.com/user-attachments/assets/966af046-a4bf-4fb8-a0e0-534af1d119ba

jonnew:
Additionally, the section controls in the channel-visualizer (probe inface GUI) are a bit clunky. Selection seems to work ok, but removal is hard. Can I make a block selection and them e.g. click or shift-click individual selected electrodes to de-select them?

bparks13:
Other potential ways to highlight where a pulse occurs that might be more performant
- Instead of N extra objects from all the individual pulses, plot only 1 extra line that does not show the actual amplitude, but rather is at a fixed height whenever each pulse occurs
    - This could be an impulse (i.e., a raster plot at the start of each pulse), or a unit step function (i.e., rectangular pulse that is high for the duration of the pulse, and low the rest of the time)
- Draw all lines that have a pulse anywhere with a different color than lines without a pulse
    - Currently the default line color is blue, and red for channels with invalid stimuli
    - We could add a different color (orange? green? I’m not sure) that we draw the waveform with when a pulse is present anywhere

bparks13:
Converted to a draft, will come back to this if needed.

PR #347, opened 2024-10-22
Rhs2116 - Search through possible step sizes to find global step size

### LINKED ISSUE #345: Instead of using the smallest step size, search for a step size that works for all stimuli
Currently, the logic is that the step size should remain the same as the current step size if possible, and if this is not possible, the new step size will be the smallest possible step size to accommodate the requested amplitude. This causes issues if the rest of the stimuli are too large to fit within the smallest step size.

Instead, we should evaluate the existing stimuli by taking their current amplitudes, and comparing them to the list of possible step sizes for the new amplitude. Search for a step size that works for all amplitudes (i.e., the steps are between 1 and 255). Choose the step size that accomplishes this.

If no such step size exists (the amplitudes are too far apart to be bridged by a common step size), give an error message immediately that the requested amplitude is outside of the bounds of the current step size. If the user continues forward, then allow them to set the new amplitude and override the existing step size, clearing the stimuli for channels that do not fit into the new step size, similar to how it is currently performed.

bparks13:
Instead of choosing the smallest step size for a new amplitude, search through all existing amplitudes and all possible step sizes to see if one fits all amplitudes simultaneously.
Modified the existing error message to be more clear why the current amplitudes are incompatible with the new step size.

- Fixes #345


OpenEphys.Onix1.Design/Rhs2116StimulusSequenceDialog.cs
```diff
@@ -1070,7 +1044,33 @@ private bool UpdateStepSizeFromAmplitude(double amplitude)
                 }
                 else
                 {
-                    StepSize = possibleStepSizes.First();
+                    // NB: Search through the possible step sizes and try to find one that matches all current amplitudes
+                    var validStepSizes = possibleStepSizes.Where(s =>
```
jonnew, line 1048: I think the crux in this process, for deciding if a stepsize is valid is the definition of `IsValidNumberOfSteps` combined with the `First()` section below. This logic indicates that any step size that allows the requested amplitude to fit can be used, without respect to how precise that fit actually is. Lets look at the space of step sizes versus requested amplitudes and the error between the requested amplitude and the one that can be actualized given a stepsize. Lets define error:

$$ error\  (percent) = 100  * \frac{A - [round(A/s) * s]}{A}$$

Where $$s$$ is the step size and $$A$$ is the requested, floating point amplitude. Lets look at the error for the $$(s, A)$$ parameter space:

<details>

<summary>Python script</summary>

``` python
import numpy as np
from matplotlib import pyplot as plt
from matplotlib import colors

step_sizes_uA = np.array([0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0])
amplitudes_uA = np.arange(0.01, 2500, 0.01)


E = np.zeros((step_sizes_uA.shape[0],amplitudes_uA.shape[0]))

for i, s in enumerate(step_sizes_uA):
    for j, a in enumerate(amplitudes_uA):
        
        if a < s or a > s * 255:
            E[i, j] = np.nan;
        else:           
            E[i, j] = 100 *(a - (round(a/s) * s)) / a
        
#%%
plt.close('all')
fig, axs = plt.subplots(1, 2, figsize=(5, 6), layout='constrained')

interp_type = 'nearest'
zoom_end = 5000

norm = colors.Normalize(vmin=np.nanmin(E), vmax=np.nanmax(E))


im = axs[1].imshow(E[:, 0:zoom_end], aspect='auto', norm=norm, interpolation=interp_type)
axs[1].set_yticks(np.arange(0, step_sizes_uA.shape[0], 1), labels=step_sizes_uA)
axs[1].set_xticks(axs[1].get_xticks(), labels=axs[1].get_xticks() * 0.01)
axs[1].set_xlabel("Amplitude [uA]")
axs[1].set_ylabel("Step size [uA]")
axs[1].set_xlim(-zoom_end/20, zoom_end+zoom_end/20)

axs[0].imshow(E, aspect='auto', norm=norm, interpolation=interp_type)
axs[0].set_yticks(np.arange(0, step_sizes_uA.shape[0], 1), labels=step_sizes_uA)
axs[0].set_xticks(axs[0].get_xticks(), labels=axs[0].get_xticks() * 0.01)
axs[0].set_xlabel("Amplitude [uA]")
axs[0].set_ylabel("Step size [uA]")
axs[0].set_xlim(-len(amplitudes_uA)/20, len(amplitudes_uA)+len(amplitudes_uA)/20)


cbar = fig.colorbar(im, ax=axs, orientation='horizontal', fraction=.1)
cbar.ax.set_xlabel("Error (%)")
```
</details>


amp-step-error

All the colored sections of this heatmap are valid step-size and amplitude combinations, but the error between the requested amplitude and the actualized one can be up to 30% for some combinations. For this reason, we must got beyond finding a "valid" step-size and find the step-size that minimizes the maximal relative error between the desired and actual amplitude across channels.

jonnew (changes requested):

OpenEphys.Onix1.Design/Rhs2116StimulusSequenceDialog.cs
```diff
@@ -927,11 +907,6 @@ private double GetTimeFromSample(uint value)
...
         }
 
-        private double GetAmplitudeFromSample(byte value)
-        {
-            return GetAmplitudeFromSample(value, StepSize);
-        }
-
         private double GetAmplitudeFromSample(byte value, Rhs2116StepSize stepSize)
         {
             return value * Rhs2116StimulusSequence.GetStepSizeuA(stepSize);
```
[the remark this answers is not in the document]
bparks13: @jonnew This feels like a big enough change to me that it should be its own issue. I've created a new one here (#353), let me know what you think of it, and if it is important to be moved to 0.4.1 we can definitely prioritize it.

bparks13:
Commenting here the algorithm used to define the step-size.

1. User types in an amplitude value [0.01 - 2550 µA]
    1. These bounds are checked to ensure it is a valid amplitude
2. A list of possible step sizes are found that can be used to generate this amplitude
    1. If the list contains the current step size that other channels use, use the current step size
        - **Note**: This currently does not evaluate error for the requested amplitude. Should the newly requested amplitude take priority over existing amplitudes? Meaning, should we change the step size if the new amplitude is slightly off, or should we prioritize the existing amplitudes and round the new amplitude to match?
    1. Otherwise, create a new list of step sizes that are valid for all existing amplitudes
        1. Using this new list, iterate over all existing and new amplitudes to calculate the error, and find a step size that minimizes the maximum error of each amplitude
    1. If no step size exists that fits all amplitudes, choose the largest step size that is valid for the new amplitude, but give a warning that the step size is out of range
        - **Note**: This does not change the step size for the existing amplitudes yet, the step size for the existing amplitudes is only updated if the user tries to apply the stimulus settings to a new channel

jonnew:
Unfortunately, this is still incorrect:

Here I'm trying to change one amplitude to 100 uA while the rest of the channels are at 1000 uA. 0.5 uA is selected and I get the following warning:

image

But both 5 uA and 10 uA steps would perfectly fit 100 and 1000 uA amplitudes.

jonnew:
I'm attaching a starting point for correcting this, potentially. I would completely forget the GUI and thing about how to modify the definitions of `Rhs2116Stimulus`, `Rhs2116StimulusSequence`, and `Rhs2116StimulusSequencePair` in order to verify a requested stimulus and optimize the step size. 

- I would make `Rhs2116StimulusSequence.StepSize` have a private setter to ensure its only going to be manipulated via internal logic
- I would make `Rhs2116StimulusSequence.Stimuli` have private setter as well. 
- I would then provide this class with a public function that can be called to try to add stimuli to the `Stimuli` while doing verification, error optimization, etc. 

``` csharp
using System;
using System.Collections;
using System.Collections.Generic;
using System.Linq;
using System.Xml.Serialization;
using System.Runtime.CompilerServices;

[assembly: InternalsVisibleTo("OpenEphys.Onix1.Design")]

namespace OpenEphys.Onix1
{
    /// <summary>
    /// A stimulus sequence for a Rhs2116 device.
    /// </summary>
    public class Rhs2116StimulusSequence
    {
        internal const int NumberOfChannels = 16;
        const int NumberOfStepSizes = 10;

        /// <summary>
        /// Initializes a new instance of the <see cref="Rhs2116StimulusSequence"/> class with 16 default
        /// stimuli.
        /// </summary>
        public Rhs2116StimulusSequence()
        {
            Stimuli = new Rhs2116Stimulus[NumberOfChannels];

            for (var i = 0; i < Stimuli.Length; i++)
            {
                Stimuli[i] = new Rhs2116Stimulus();
            }
        }

        /// <summary>
        /// Initializes a new instance of the <see cref="Rhs2116StimulusSequence"/> by performing a
        /// shallow copy of a reference sequence.
        /// </summary>
        /// <param name="sequence">Existing <see cref="Rhs2116StimulusSequence"/> to copy.</param>
        public Rhs2116StimulusSequence(Rhs2116StimulusSequence sequence)
        {
            Stimuli = Array.ConvertAll(sequence.Stimuli, stimulus => stimulus.Clone());
            CurrentStepSize = sequence.CurrentStepSize;
        }


        public bool DefineStimulus(int channel, uint numberOfStimuli, bool anodicFirst, double delayMillis, double dwellMillis, 
            double anodicMillis, double anodicMicroAmps, double cathodicMillis, double cathodicMicroAmps, double interStimMillis,
            bool force)
        {
            // TODO: sanity checks other than amplitude checking
            var error = new double[NumberOfChannels * 2, NumberOfStepSizes];
            var maxError = new List<double>(NumberOfStepSizes);
            var stepSizes = Enum.GetValues(typeof(Rhs2116StepSize)).Cast<Rhs2116StepSize>();
            var stepSizesuA = stepSizes.Select(s => GetStepSizeuA(s)).ToArray();
            var currentStepSizeuA = GetStepSizeuA(CurrentStepSize);

            for (int c = 0; c < NumberOfChannels; c += 2)
            {
                var anodicAmp = c == channel ? anodicMicroAmps : Stimuli[c].AnodicAmplitudeSteps * currentStepSizeuA;
                var cathodicAmp = c == channel ? cathodicMicroAmps : Stimuli[c].CathodicAmplitudeSteps * currentStepSizeuA;

                for (int s = 0; s < NumberOfStepSizes; s++)
                {
                    error[c, s] = anodicAmp < stepSizesuA[s] || anodicAmp > s * 255 ?
                        double.PositiveInfinity :
                        Math.Abs((anodicAmp - (stepSizesuA[s] * Math.Round(anodicAmp / stepSizesuA[s]))) / anodicAmp);

                    error[c + 1, s] = cathodicAmp < stepSizesuA[s] || cathodicAmp > s * 255 ?
                        double.PositiveInfinity :
                        Math.Abs((cathodicAmp - (stepSizesuA[s] * Math.Round(cathodicAmp / stepSizesuA[s]))) / cathodicAmp);

                    maxError[s] = Math.Max(maxError[s], Math.Max(error[c, s], error[c + 1, s]));

                }
            }

            var optimalStep = maxError.ToList().IndexOf(maxError.Min());
            var optimalError = maxError[optimalStep];

            if (optimalError == double.PositiveInfinity && !force)
                return false;

            CurrentStepSize = stepSizes.ElementAt(optimalStep);
            // TODO: regenerate Stimuli by assigning Stimuli[channel] and updated the others' parameters with the new step size

            return true;
        }

        /// <summary>
        /// Gets or sets the array of <see cref="Rhs2116Stimulus"/> sequences.
        /// </summary>
        public Rhs2116Stimulus[] Stimuli { get; private set; }

        /// <summary>
        /// Gets or sets the <see cref="Rhs2116StepSize"/>. 
        /// </summary>
        public Rhs2116StepSize CurrentStepSize { get; private set; } = Rhs2116StepSize.Step5000nA;

        /// <summary>
        /// Maximum length of the sequence across all channels.
        /// </summary>
        [XmlIgnore]
        public uint SequenceLengthSamples
        {
            get
            {
                uint max = 0;

                foreach (var stim in Stimuli)
                {
                    var len = stim.DelaySamples + stim.NumberOfStimuli * (stim.AnodicWidthSamples + stim.CathodicWidthSamples + stim.DwellSamples + stim.InterStimulusIntervalSamples);
                    max = len > max ? len : max;

                }

                return max;
            }
        }

        /// <summary>
        /// Maximum peak to peak amplitude of the sequence across all channels.
        /// </summary>
        [XmlIgnore]
        public int MaximumPeakToPeakAmplitudeSteps
        {
            get
            {
                int max = 0;

                foreach (var stim in Stimuli)
                {
                    var p2p = stim.CathodicAmplitudeSteps + stim.AnodicAmplitudeSteps;
                    max = p2p > max ? p2p : max;

                }

                return max;
            }
        }

        /// <summary>
        /// Is the stimulus sequence well defined.
        /// </summary>
        [XmlIgnore]
        public bool Valid => Stimuli.ToList().All(s => s.Valid);

        /// <summary>
        /// Does the sequence fit in hardware.
        /// </summary>
        [XmlIgnore]
        public bool FitsInHardware => StimulusSlotsRequired <= Rhs2116.StimMemorySlotsAvailable;

        /// <summary>
        /// The maximum number of memory slots available.
        /// </summary>
        [XmlIgnore]
        public int MaxMemorySlotsAvailable => Rhs2116.StimMemorySlotsAvailable;

        /// <summary>
        /// Number of hardware memory slots required by the sequence.
        /// </summary>
        [XmlIgnore]
        public int StimulusSlotsRequired => DeltaTable.Count;

        /// <summary>
        /// Gets the current step size in µA.
        /// </summary>
        [XmlIgnore]
        public double CurrentStepSizeuA
        {
            get => GetStepSizeuA(CurrentStepSize);
        }

        /// <summary>
        /// Gets the step size in microamps of a given <see cref="Rhs2116StepSize"/>.
        /// </summary>
        /// <param name="stepSize">The <see cref="Rhs2116StepSize"/> value to convert to microamps.</param>
        /// <returns>Returns a double corresponding to the given step size in microamps.</returns>
        /// <exception cref="ArgumentException">Invalid stimulus step size selection.</exception>
        public static double GetStepSizeuA(Rhs2116StepSize stepSize)
        {
            return stepSize switch
            {
                Rhs2116StepSize.Step10nA => 0.01,
                Rhs2116StepSize.Step20nA => 0.02,
                Rhs2116StepSize.Step50nA => 0.05,
                Rhs2116StepSize.Step100nA => 0.1,
                Rhs2116StepSize.Step200nA => 0.2,
                Rhs2116StepSize.Step500nA => 0.5,
                Rhs2116StepSize.Step1000nA => 1.0,
                Rhs2116StepSize.Step2000nA => 2.0,
                Rhs2116StepSize.Step5000nA => 5.0,
                Rhs2116StepSize.Step10000nA => 10.0,
                _ => throw new ArgumentException("Invalid stimulus step size selection."),
            };
        }

        /// <summary>
        /// Gets the maximum possible amplitude for a single pulse, in µA.
        /// </summary>
        [XmlIgnore]
        public double MaxPossibleAmplitudePerPhaseMicroAmps => CurrentStepSizeuA * 255;

        internal IEnumerable<byte> AnodicAmplitudes => Stimuli.ToList().Select(x => x.AnodicAmplitudeSteps);

        internal IEnumerable<byte> CathodicAmplitudes => Stimuli.ToList().Select(x => x.CathodicAmplitudeSteps);

        /// <summary>
        /// Generate the delta-table representation of this stimulus sequence that can be uploaded to the
        /// RHS2116 device. The resultant dictionary has a time, in samples, as the key and a combined
        /// [polarity, enable] bit field as the value.
        /// </summary>
        [XmlIgnore]
        internal Dictionary<uint, uint> DeltaTable
        {
            get => GetDeltaTable();
        }

        private Dictionary<uint, uint> GetDeltaTable()
        {
            var table = new Dictionary<uint, BitArray>();

            for (int i = 0; i < NumberOfChannels; i++)
            {
                var s = Stimuli[i];

                var e0 = s.AnodicFirst ? s.AnodicAmplitudeSteps > 0 : s.CathodicAmplitudeSteps > 0;
                var e1 = s.AnodicFirst ? s.CathodicAmplitudeSteps > 0 : s.AnodicAmplitudeSteps > 0;
                var d0 = s.AnodicFirst ? s.AnodicWidthSamples : s.CathodicWidthSamples;
                var d1 = d0 + s.DwellSamples;
                var d2 = d1 + (s.AnodicFirst ? s.CathodicWidthSamples : s.AnodicWidthSamples);

                var t0 = s.DelaySamples;

                for (int j = 0; j < s.NumberOfStimuli; j++)
                {
                    AddOrInsert(ref table, i, t0, s.AnodicFirst, e0);
                    AddOrInsert(ref table, i, t0 + d0, s.AnodicFirst, false);
                    AddOrInsert(ref table, i, t0 + d1, !s.AnodicFirst, e1);
                    AddOrInsert(ref table, i, t0 + d2, !s.AnodicFirst, false);

                    t0 += d2 + s.InterStimulusIntervalSamples;
                }
            }

            return table.ToDictionary(d => d.Key, d =>
            {
                int[] i = new int[1];
                d.Value.CopyTo(i, 0);
                return (uint)i[0];
            });
        }

        private static void AddOrInsert(ref Dictionary<uint, BitArray> table, int channel, uint key, bool polarity, bool enable)
        {
            if (table.ContainsKey(key))
            {
                table[key][channel] = enable;
                table[key][channel + 16] = polarity;
            }
            else
            {
                table.Add(key, new BitArray(32, false));
                table[key][channel] = enable;
                table[key][channel + 16] = polarity;
            }
        }
    }
}


```

OpenEphys.Onix1/Rhs2116Stimulus.cs
```diff
@@ -8,6 +9,90 @@ namespace OpenEphys.Onix1
     /// </summary>
     public class Rhs2116Stimulus
     {
+        /// <summary>
+        /// Double defining the length of a single time sample, in milliseconds.
+        /// </summary>
+        public static readonly double SamplePeriodMilliSeconds = 1e3 / 30.1932367151e3;
```
jonnew, line 15: Redundant with ConfigureRhs2116.SampleFrequencyHz. If you want a milliseconds version, they use that const to make it. Additionally:

1. Why not const?
2. Why declared here instead of in design library since thats the only place its used?

OpenEphys.Onix1/Rhs2116Stimulus.cs
```diff
@@ -8,6 +9,90 @@ namespace OpenEphys.Onix1
...
+            DelaySamples = delaySamples;
+            DwellSamples = dwellSamples;
+            AnodicAmplitudeSteps = anodicAmplitudeSteps;
+            AnodicWidthSamples = anodicWidthSamples;
+            CathodicAmplitudeSteps = cathodicAmplitudeSteps;
+            CathodicWidthSamples = cathodicWidthSamples;
+            InterStimulusIntervalSamples = interStimIntervalSamples;
+        }
+
+        internal void CopyStimulus(Rhs2116Stimulus stimulus)
```
jonnew, line 65: You have a copy constructor above. What is the purpose of this?

jonnew, line 65: Additionally, `objects` com pre-packaged with a lot of functionality like this

https://learn.microsoft.com/en-us/dotnet/api/system.object.memberwiseclone?view=net-8.0

but this can also be used for deep copies

```
    public Person DeepCopy()
    {
        Person other = (Person)MemberwiseClone();
        other.IdInfo = new IdInfo(IdInfo.IdNumber);
        return other;
    }
```

OpenEphys.Onix1/Rhs2116Stimulus.cs
```diff
@@ -8,6 +9,90 @@ namespace OpenEphys.Onix1
...
+        }
+
+        /// <summary>
+        /// Requested amplitude of the anodic pulse in microamps
+        /// </summary>
+        [DisplayName("Requested Anodic Amplitude (µA)")]
+        [XmlIgnore]
+        [Browsable(false)]
+        [Externalizable(false)]
+        public double RequestedAnodicAmplitudeuA { get; set; } = 0.0;
```
jonnew, line 85: This is not a property of a `Rhs2116Stimulus`, which is a representation of a stimulus that will be put into hardware. Therefore, this class should not contain this property.

OpenEphys.Onix1/Rhs2116StimulusSequencePair.cs
```diff
@@ -9,8 +12,8 @@ namespace OpenEphys.Onix1
     /// </summary>
     public class Rhs2116StimulusSequencePair
     {
-        internal readonly Rhs2116StimulusSequence StimulusSequenceA;
-        internal readonly Rhs2116StimulusSequence StimulusSequenceB;
+        internal virtual Rhs2116StimulusSequence StimulusSequenceA { get; }
```
jonnew, line 15: I'm confused as to why this is marked virtual. Are you planning on overriding these properties?

OpenEphys.Onix1/Rhs2116StimulusSequencePair.cs
```diff
@@ -9,8 +12,8 @@ namespace OpenEphys.Onix1
...
+        internal virtual Rhs2116StimulusSequence StimulusSequenceB { get; }
```
jonnew, line 16: why virtual?

OpenEphys.Onix1/Rhs2116StimulusSequencePair.cs
```diff
@@ -33,10 +36,22 @@ public Rhs2116StimulusSequencePair(Rhs2116StimulusSequencePair stimulusSequenceD
...
+        {
+            StimulusSequenceA = new Rhs2116StimulusSequence(stimulusSequenceA);
+            StimulusSequenceB = new Rhs2116StimulusSequence(stimulusSequenceB);
+        }
+
         /// <summary>
         /// Gets or sets the <see cref="Rhs2116Stimulus"/> array of stimuli.
         /// </summary>
-        public Rhs2116Stimulus[] Stimuli
+        public virtual Rhs2116Stimulus[] Stimuli
```
jonnew, line 54: why virtual?

jonnew (changes requested):
I terms of behavior, this looks very good. I played around with several extreme combinations of amplitudes

- At margins of min and max for an step size i had in mind
- Looking at the rounding behavior versus actual applied amplitude in grey box
- Looking at display while i was doing this.

All of that looked good. However, the underlying code does need work.

OpenEphys.Onix1/Rhs2116Stimulus.cs
jonnew: What I mean by this is the following: when you are defining a class, the better way to think about it is that you are defining a type. A type indicates a set of possible features that belong and a set that dont. In this case, in my opinion, these properties do not belong to a software representation of a stimulus because they have to do with a user's desires rather than what is possible.  They should be held somewhere else, probably in the design library.

OpenEphys.Onix1/Rhs2116Stimulus.cs
bparks13: Great point, I'm not sure why I added this instead of using the copy constructor. Removed this method.

bparks13:
@jonnew I have addressed the review comments. Let me know if this PR is approved, and if it is I will go ahead and rebase to clean up the history before merging

OpenEphys.Onix1.Design/Rhs2116StimulusSequenceDialog.cs
```diff
@@ -12,10 +12,18 @@ namespace OpenEphys.Onix1.Design
     /// </summary>
     public partial class Rhs2116StimulusSequenceDialog : Form
     {
+        /// <summary>
```
jonnew, line 15: For the sake of consistency, i would remove this from a private variable (imagine if they all had something like this).

OpenEphys.Onix1/Rhs2116Stimulus.cs
```diff
@@ -212,7 +150,7 @@ public bool IsValid(out string reasonInvalid)
         /// <summary>
         /// Resets all properties to their default values.
         /// </summary>
-        public void Clear(bool clearRequestedAmplitudes = false)
+        public void Clear()
```
jonnew, line 153: et voilà -- we can see an effect of type pollution cleanup here

jonnew (approved):
Looks good. Can we add the following functionality before the merge please:

image

Issue #313, opened 2024-09-26
Save zoom state when resizing dialogs


bparks13:
If a dialog is resized (maximizing the window for example), it resets the zoom state to be fully zoomed out. This is potentially annoying, but not a critical feature to fix.

Should be a simple matter of grabbing the current axis boundaries during the Resize event, and applying them after the resizing is complete.

bparks13:
Example GIF of the behavior in question.

![Image](https://github.com/user-attachments/assets/c04e1fc8-dfcb-43a2-ac18-7a085fb7fb5d)

Note that this does not affect the Rhs2116 stimulus sequence dialog, but it does affect the `ChannelConfigurationDialog`.

PR #348, opened 2024-10-22
Allow individual parameters to be changed

### LINKED ISSUE #344: Rhs2116 Stimulus Sequence Dialog - Change individual parameters across all channels
Right now the GUI allows you to select multiple channels and use the Define Stimulus section to apply stimuli to the selected channels:

image

But what if the user wants to edit one feature of the stimuli (e.g. amplitude or pulse width) but leave all other characteristics unchanged (e.g. the times of pulses). There should be some way to allow this.

bparks13:
Instead of requiring all parameters to be filled out and then subsequently written to a channel, this update now allows only those parameters that are filled in to be applied to a channel or channels.

For example, instead of having to fill out all timings and amplitudes, if the pulse width is the only parameter that needs to be updated, the pulse width is the only variable that is filled in when `Apply` is pressed.

Added additional checks if the pulse is biphasic symmetric to ensure consistency.

- Fixes #344

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <PackageIcon>icon.png</PackageIcon>
-    <VersionPrefix>0.4.0</VersionPrefix>
+    <VersionPrefix>0.4.1</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <LangVersion>10.0</LangVersion>
     <Features>strict</Features>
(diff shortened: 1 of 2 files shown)


jonnew:
This looks really good. I like the behavior a lot. The only critique I have is that the plot zoom resets when parameters are changed. The reason this is a bit annoying is because for a sequence I'm zoomed in to the portion of the waveform that I want to change. When I go edit e.g. an amplitude parameter and expect to see the at change reflected in the waveform, the view resizes and I have manually zoom in again to see if my change had the indented effect.

If there is a way to save the zoom state and reapply when parameters are changed, I would like that. Otherwise, this looks very good.

bparks13:
@jonnew I have created a new issue regarding saving the zoom state (#349), feel free to add any more thoughts to the issue. For now, I have set it as the 0.5.0 milestone, but we can push for it to be 0.4.1 if we fully define the problem quickly. 

There is a very related issue (#313) that can also be fleshed out at the same time related to resizing dialogs.

jonnew (approved):

PR #350, opened 2024-10-28
Update readme and URLs

cjsha:
The important thing from this PR is that we need to update the URLs (after the repo name changed) so that the nodes that are published on nuget are linked to the actual docs site. Currently the "F1" Help functionality doesn't work for our nodes. I think this should fix that.

I also added some instructions to the README to document how to setup breakpoints in Visual Studio.


README.md
```diff
@@ -18,4 +18,17 @@ Acquisition System](https://open-ephys.github.io/onix-docs).
    - Run the Bonsai application installed in step 1
    - Instruct Bonsai to load the compiled library
 
-<img alt="Select which library to compile" src="./.bonsai/build-and-run.png" width="60%" />
+<img alt="Select which library to compile" src="./images/build-and-run.png" width="60%" />
+
+### Enable breakpoints
+
```
jonnew, line 24: If you are going to provide information on this, then you have to also provide instructions in installing the child process debugging extension in VS:

image

README.md
```diff
@@ -18,4 +18,17 @@ Acquisition System](https://open-ephys.github.io/onix-docs).
...
+1. Enable child process debugging
+   -  Navigate to child process debugging settings
+      ![Navigate to child process menu](./images/enable-child-process-navigation.png)
+   -  Confirm the following option is checked: 
+      ![Enable child process](./images/enable-child-process-checkbox.png)
+1. Enable mixed debugging
```
jonnew, line 30: mixed debugging -> native code debugging

jonnew (changes requested):
Instead of random snippets with no context:

image

image

Just show the compete dialog with steps labelled as in the original image in this readme, or no labels if its obvious

bparks13 (changes requested):
I addressed all of the comments by @jonnew, and I updated the version number so that the actions would pass.

However, I noticed that in the `build-and-run.png` file, there is only the numbers for 3 and 4. Did we change the image?

cjsha:
I haven't changed `build-and-run.png`. I just moved it to a different images folder. In any case, I think the way it is makes sense - labels 3 & 4 correspond to bullet points 3 & 4 of the "Run in visual studio" section. I'm not sure bullet points 1 & 2 need screenshots w/ labels.

I don't think we had it like jon does in the "[Enable Debugging](https://github.com/open-ephys/bonsai-onix1/tree/update-README-and-URLs?tab=readme-ov-file#enable-debugging)" section that we'd created. In the latest commit, I try to amend that.

Thanks for addressing his comments @bparks13

bparks13 (approved):
Everything looks good to me now, but there are a few commits going back and forth; before merging to main, I would recommend someone go through and squash commits together to clean up the history

jonnew (approved):

jonnew:
@cjsha please squash and merge

PR #359, opened 2024-10-29
Hardcode ConfigureHeartbeat to enabled in ConfigureBreakoutBoard

### LINKED ISSUE #352: Hardcode heartbeat enable property to true in ConfigureBreakoutBoard
This should be done by firmware as noted here:

https://github.com/open-ephys/onix-pcie-host-hdl/issues/4

However, we can at least get rid of this the enable field in `ConfigureHeartbeat` here:

https://github.com/open-ephys/bonsai-onix1/blob/18dd66bcf586edc00506209491035ecce7142739/OpenEphys.Onix1/ConfigureBreakoutBoard.cs#L31C61-L31C67

By using a similar strategy to the `AxisMap` properties of `ConfigureBno055`:

https://github.com/open-ephys/bonsai-onix1/blob/18dd66bcf586edc00506209491035ecce7142739/OpenEphys.Onix1/ConfigureNeuropixelsV2eHeadstage.cs#L51

jonnew:
- Additionally, hardcode its BeatsPerSecond to 100
- Fixes #352

### DIFF
--- a/OpenEphys.Onix1/ConfigureBreakoutBoard.cs
+++ b/OpenEphys.Onix1/ConfigureBreakoutBoard.cs
@@ -25,10 +25,13 @@ namespace OpenEphys.Onix1
         /// <summary>
         /// Gets or sets the heartbeat configuration.
         /// </summary>
-        [TypeConverter(typeof(SingleDeviceFactoryConverter))]
+        /// <remarks>
+        /// This heartbeat is always enabled and beats at 100 Hz.
+        /// </remarks>
+        [TypeConverter(typeof(HeartbeatSingleDeviceFactoryConverter))]
         [Description("Specifies the configuration for the heartbeat device in the ONIX breakout board.")]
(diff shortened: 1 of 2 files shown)


cjsha (approved):
I'm not immediately qualified to evaluate the code as I'm not as familiar with OpenEphys.Onix1. So, I will do this review from the perspective of a user.

I confirm that there is a 100Hz heartbeat that cannot be disabled and its frequency cannot be changed.

Is it worth considering removing it from the properties pane completely? Its label is there with no properties underneath. 
image

It seems like Heartbeat is better as something that should only appear to the user only if the user needs to use it like in a troubleshooting tutorial or something like that. In any case, this is a nitpick, and I will approve the PR anyway.

jonnew:
I think that if memory monitor is there, then heartbeat can be there. I don't really like hiding things. I think its fine.

PR #362, opened 2024-10-31
Simplify PortStatusFrame

### LINKED ISSUE #351: Include Lock and Pass `PortStatusCode` enum
The `PortStatusCode` is already a `[Flags]` `enum`

https://github.com/open-ephys/bonsai-onix1/blob/18dd66bcf586edc00506209491035ecce7142739/OpenEphys.Onix1/ConfigurePortController.cs#L120

However, port `PortStatusDataFrame` includes SERDES Lock and Pass as independent bools. 

There is no reason I can see for not packing these bools into the `PortStatusCode` `enum` as a couple additional Flag values. The reason this was done originally is that the way its done now is more straight reflection of the hardware data frame, but why do we need to respect that at this level?

jonnew:
- Instead of making SerdesPass and SerdesLock bools next to the code word, the a codeword valid flag required to indicate if the codeword should be none, jusst make the PortStatusCode flag enum 16-bits long and contain status code, lock, and pass status together
- Needs to be tested with hardware
- Code valid flag is currently not interpreted by PortStatusCode since its not relevant for this interpretation of the low level frame data, but I need to confirm that this is not going to do something strange to the StatusCode or result in exception
- Fixes #351

### DIFF
--- a/OpenEphys.Onix1/ConfigurePortController.cs
+++ b/OpenEphys.Onix1/ConfigurePortController.cs
@@ -53,7 +53,7 @@ namespace OpenEphys.Onix1
                 context.HubState = (PassthroughState)(((int)context.HubState & ~(1 << portShift)) | passthroughState);
 
                 // leave in standard mode when finished
-                return Disposable.Create(() => context.HubState = (PassthroughState)((int)context.HubState & ~(1 << portShift))); 
+                return Disposable.Create(() => context.HubState = (PassthroughState)((int)context.HubState & ~(1 << portShift)));
             })
             .ConfigureLink(context =>
             {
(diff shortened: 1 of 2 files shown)


aacuevas (commented):
The port status device defines a code with the 0x1000 flag, for crc error on init. While I do not believe this can actually reach the output, should it not be safer to add it to the enum?

OpenEphys.Onix1/ConfigurePortController.cs
```diff
@@ -117,36 +117,45 @@ internal enum HubConfiguration
...
     {
         /// <summary>
-        /// Specifies that the status code should be disregarded.
+        /// Specifies the SERDES forward channel lock status.
         /// </summary>
-        Invalid = 0x0,
+        SerdesLock = 0x0001,
+
+        /// <summary>
+        /// Gets the SERDES on-chip parity check status.
```
bparks13, line 128: This deviates from the pattern of the other summaries. Should it be changed to "Specifies the SERDES on-chip parity check status."?

aacuevas (approved):

PR #369, opened 2024-11-08
Hide `PolledBno055` properties when embedded in a headstage

### LINKED ISSUE #314: Hide `PolledBno055` properties in GUI dialogs
When a `PolledBno055` is under a headstage operator, the `AxisMap` and `AxisSign` variables are not displayed in the Bonsai property pane because these are fixed quantities when paired with a specific headstage. However, when they are added to the default property pane in a configuration GUI, these properties are now shown and editable.

image

These variables are not copied back into the main Bonsai window, so there will not be any permanent changes, but if a user changes these values and expects them to make a difference it will seem like a bug.

bparks13:
Utilize the same type converter from the main library to hide `AxisMap` and `AxisSign` from the property grid when opened in dialogs. This is accomplished by creating an internal class that is only used to wrap the `ConfigurePolledBno055` class in a public get / set so that it acts like an expandable object.

These hidden properties are editable from the `ConfigurePolledBno055` operator, since right now I don't know how to distinguish a truly standalone `PolledBno055` from one that is part of a headstage, unless it is called from the headstage editor (i.e., `ConfigureNeuropixelsV1eHeadstage`). 

Fixes #314

### DIFF
--- a/OpenEphys.Onix1.Design/NeuropixelsV1eHeadstageEditor.cs
+++ b/OpenEphys.Onix1.Design/NeuropixelsV1eHeadstageEditor.cs
@@ -23,7 +23,7 @@ namespace OpenEphys.Onix1.Design
 
                     if (editorDialog.ShowDialog() == DialogResult.OK)
                     {
-                        configureHeadstage.Bno055.Enable = editorDialog.DialogBno055.ConfigureNode.Enable;
+                        configureHeadstage.Bno055.Enable = editorDialog.DialogBno055.Bno055.Enable;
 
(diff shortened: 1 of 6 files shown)


jonnew:
As of right now, I'm very hesitant to merge this. This is a pretty hard issue, it turns out, but the precedent that this solution sets is not good. It requires more logic being created on the design side of things than was required to hide the property in the base library. I think we need to explore other options.

bparks13:
Closing this for now, we have not come up with a better solution so for now it will be left here until we can come back to it.

PR #389, opened 2024-12-18
Fix ConfigurePolledBno055 operation mode

### LINKED ISSUE #378: Investigate hs64 bno orientation
From https://github.com/open-ephys/bonsai-onix1-docs/issues/147:

> - the imu axes are wrong, should be 0, -1, 0 for proper operation

If this is true, I suspect it should be fixed in the source code, not a docs issue.

### LINKED ISSUE #388: ConfigurePolledBno055 does not configure the BNO055 in NDOF mode
This line configures the Bno055 in IMU mode

https://github.com/open-ephys/bonsai-onix1/blob/74f6f22d3df6ae18b9f7deccc57923cf5eaf2f13/OpenEphys.Onix1/ConfigurePolledBno055.cs#L122

This is almost certainly what is responsible for the large drift people are seeing with the neuropixels headstages. It should be 0x0C.

jonnew:
- It was incorrectly configured in IMU mode (which only uses gyroscope and accelerometer). Now its correctly configured to use magnetometer in NDOF mode, which should reduce drift.
- Fixes #388 
- Fixes #378

### DIFF
--- a/OpenEphys.Onix1/ConfigurePolledBno055.cs
+++ b/OpenEphys.Onix1/ConfigurePolledBno055.cs
@@ -119,7 +119,7 @@ namespace OpenEphys.Onix1
             i2c.WriteByte(0x3F, 0x00); // Internal oscillator
             i2c.WriteByte(0x41, (uint)AxisMap);  // Axis map config
             i2c.WriteByte(0x42, (uint)AxisSign); // Axis sign
-            i2c.WriteByte(0x3D, 0x0C); // Operation mode is NOF
+            i2c.WriteByte(0x3D, 0x0C); // Operation mode is NDOF
         }
     }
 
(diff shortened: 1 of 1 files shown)


OpenEphys.Onix1/ConfigurePolledBno055.cs
```diff
@@ -119,7 +119,7 @@ void ConfigureBno055(DeviceContext device)
             i2c.WriteByte(0x3F, 0x00); // Internal oscillator
             i2c.WriteByte(0x41, (uint)AxisMap);  // Axis map config
             i2c.WriteByte(0x42, (uint)AxisSign); // Axis sign
-            i2c.WriteByte(0x3D, 8); // Operation mode is NOF
+            i2c.WriteByte(0x3D, 0x0C); // Operation mode is NOF
```
bparks13, line 122: The PR comments state the mode is "NDOF", but the comment here states "NOF". Is this correct?

cjsha: It's for sure NDOF

PR #391, opened 2024-12-18
Explicitly define orientation of Rhs2116 channel configuration

### LINKED ISSUE #387: Indicate orientation of RHS2116 PCB in the channel selection window of the RHS2116 GUI
This issue stems from message in the onix-docs channel Dec 12/13th.

bparks13:
Added contact annotation labels to the default configuration that aligns with documentation images (i.e., A1 / B1 / etc.).
Added a label to the top of the image that defines the side of the PCB that contains the tether.

- Fixes #387

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <PackageIcon>icon.png</PackageIcon>
-    <VersionPrefix>0.4.2</VersionPrefix>
+    <VersionPrefix>0.4.3</VersionPrefix>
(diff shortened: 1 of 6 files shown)


cjsha (approved):
looks good to me 
image

bparks13:
Another zoomed in image for reference:
image

ChucklesOnGitHub:
Hmm, I'm not sure I follow the logic of the numbering.

Using this mapping:
image

If you look at the headstage straight on (looking at the probe connector prongs) with the tether connector on top, then I think the Intan channel numbers would be:
image

jonnew:
Its pretty obvious to me that this cartoon depicts looking directly input the business end of the headstage. But, there actually nothing that shows that explicitly here. Probably not worth worrying about.

PR #393, opened 2024-12-23
Rhs2116 step size algorithm now accounts for requested amplitude

### LINKED ISSUE #392: Rhs2116 step size algorithm does not account for requested amplitude
The automated algorithm for choosing the step size with the minimum error does not account for the requested amplitude; it only uses the existing stimuli. This means that if there are no amplitudes stored in the `Stimuli` array, there are a number of cases where the step size that is chosen does not minimize the error.

As a concrete example, when opening a new configuration window the default step size is `Rhs2116StepSize.Step5000nA` (5 µA), and the stimuli array is empty. If the user then requests 0.12 µA, the algorithm will determine that there are no differences between any of the possible step sizes, and chooses the one closest to the original step size to minimize error. This leads to `Rhs2116StepSize.Step100nA` (0.1 µA) being chosen as the new step size. However, this means that there is an error of 0.02 µA between the requested amplitude and the actual value that will be used for stimulation.

There is a current workaround for this, but it requires the user to write 0.01 as the requested amplitude and apply it to a random channel so that it sets the current step size as `0.01 µA` before they can then write in 0.12 and have it applied correctly.

To solve:
Add the requested amplitude as an input to the internal method that algorithmically determines the new step size. This will ensure that even if there are no stimuli saved, it will use this amplitude as a reference to minimize the error.

Addendum: Should we change the default step size to be `Rhs2116StepSize.Step10nA` for new objects? This does not solve the current problem, but it might make more sense to start at the highest resolution step size and then go up if the user requires a higher amplitude than this can achieve.

bparks13:
When calculating the error to find the optimum step size, the algorithm now utilizes the requested amplitude so that in the case when no stimuli are present, the error is correctly modified for each step size and the appropriate size is chosen for the requested amplitude

- Fixes #392

### DIFF
--- a/OpenEphys.Onix1/Rhs2116StimulusSequence.cs
+++ b/OpenEphys.Onix1/Rhs2116StimulusSequence.cs
@@ -230,7 +230,7 @@ namespace OpenEphys.Onix1
             }
         }
 
-        internal static Rhs2116StepSize GetStepSizeWithMinError(IEnumerable<Rhs2116StepSize> stepSizes, Rhs2116Stimulus[] stimuli, Rhs2116StepSize currentStepSize)
+        internal static Rhs2116StepSize GetStepSizeWithMinError(IEnumerable<Rhs2116StepSize> stepSizes, Rhs2116Stimulus[] stimuli, double requestedAmplitude, Rhs2116StepSize currentStepSize)
         {
             var numberOfStepSizes = stepSizes.Count();
             var maxError = new List<double>(numberOfStepSizes);
@@ -263,6 +263,12 @@ namespace OpenEphys.Onix1
 
                     maxError[s] = Math.Max(maxError[s], Math.Max(anodicError, cathodicError));
                 }
+
+                var requestedError = requestedAmplitude < stepSizesuA[s] || requestedAmplitude > stepSizesuA[s] * 255 ?
+                        double.PositiveInfinity :
+                        CalculateError(requestedAmplitude, stepSizesuA[s]);
+
+                maxError[s] = Math.Max(maxError[s], requestedError);
             }
 
             if (maxError.Distinct().Count() == 1)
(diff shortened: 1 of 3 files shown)


OpenEphys.Onix1/Rhs2116StimulusSequence.cs
```diff
@@ -263,6 +263,12 @@ static double CalculateError(double amplitude, double stepSizeuA)
 
                     maxError[s] = Math.Max(maxError[s], Math.Max(anodicError, cathodicError));
                 }
+
+                var requestedError = requestedAmplitude < stepSizesuA[s] || requestedAmplitude > stepSizesuA[s] * 255 ?
```
jonnew, line 267: I wonder if these checks are needed, or need to be modified:

- Consider step size  is 10 nA and requested amplitude is 9.99999 nA.
- Consider step size is 10 nA and requested amplitude is 2550.00001 nA. 

It seems like the logic can be simplified such that an infinite error is returned from `CalculateError` if the following evaluates to true:


```
Math.Round(amplitude / stepSizeuA) > 255 || Math.Round(amplitude / stepSizeuA) == 0
```

and then all these ternary operators can be removed.

reults in 0 or

bparks13: @jonnew Currently the GUI does not allow the user to input those values you listed. I put a hard check there which forces the value to be `10 nA <= value <= 2550 µA`. This was done because those are the limits placed by the hardware, so it made sense to me that the user should never be able to request a value outside of those limits. However, I see what you mean about checking for rounding at the edges of the limit.

The question then becomes should there be any sort of boundary checking when the user types in a value? I think that requiring an amplitude > 0 is a valid lower limit, but I'm not sure about the upper limit.

jonnew: The key here is that the boundary checking should be done in one place. RN its spread to the control input, and all these ternary operations. The later of which don't actually provide a truly correct answer because there are situations close to the edge of a single or 255 steps where they exclude it. 

I don't think you need to remove the global 10 nA to 2.55 mA check at the control level. Inside this function, i would take the strategy I'm suggesting where the check is done implicitly using the logic proposed above inside the Calculate error. That way its done in one place and without the edge cases.

jonnew (approved):
These changes seem to work well. I added some comments to the intial issue, but they probably warrant a separate PR, so Im going to approve this.

PR #398, opened 2025-01-13
Fix axis map on headtages that use PolledBno055

### LINKED ISSUE #396: Verify Bno055 axis map functionality in NDOF mode
There is some ambiguity as to what, exactly, the AXIS_MAP and AXIS_SIGN registers do in the context of NDOF mode for the BNO055. The datasheet claims NDOF is absolute relative to the Earth's B field and Gravity vector. So what does remapping axes do in this situation? Is there any way to make this measurement relative to a tether orientation? We need to make some measurements in different axis maps to see how data changes.

At the end of the day, what we are trying to figure out is where we need to put a selector for determining the "tether axis" for commutation. If we are "stuck" with everything being relative the B field G-vector, in some sense, this is good because its very consistent across headstages (assuming that calibration is good). However, it provides no information about how untwisting should be done.

jonnew:
- Fix XML documentation of various properties of Bno055 data and configuration to reflect our understanding of the chips operation
- Fixes #396

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <PackageIcon>icon.png</PackageIcon>
-    <VersionPrefix>0.4.2</VersionPrefix>
+    <VersionPrefix>0.4.3</VersionPrefix>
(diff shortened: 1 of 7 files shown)


bparks13 (approved):
Looks good, I went ahead and bumped the version number so this PR could be merged, and changed a few words to make it flow better.

PR #403, opened 2025-01-17
Add missing configuration to ConfigureUCLAMiniscopeV4Camera

### LINKED ISSUE #402: Miniscope not sending camera data, but can be controlled and IMU sensor data acquired
The camera is not sending data, but the quaternion and setting LED and focus work. The scope is working correctly with the miniscope-daq library. The device is enabled. Workflow attached.

Tried with different tethers and voltages:
With micro-coax, 3.5 V at the scope with 5.7 as the PortVoltage (max override I can set). 
With mini-coax, 4V with autodetect and 4.4V with 5.7 as PortVoltage.

Image

[test_miniscope.zip](https://github.com/user-attachments/files/18457274/test_miniscope.zip)

jonnew:
- The ConfigureUCLAMiniscopeV4Camera passthrough configuration assumed that passthrough device registers were defaulted prior to configuring the device to receive miniscope data. If the passthrough device was used for something else (e.g. neuropixels headstage) before this, then this would not be the case. Now we explicitly set all passthrough registers in ConfigureUCLAMiniscopeV4Camera to the correct values.
- Fixes #402

### DIFF
--- a/OpenEphys.Onix1/ConfigureUclaMiniscopeV4Camera.cs
+++ b/OpenEphys.Onix1/ConfigureUclaMiniscopeV4Camera.cs
@@ -191,6 +191,7 @@ namespace OpenEphys.Onix1
 
             // The camera does not rely on magic words for data alignment
             device.WriteRegister(DS90UB9x.MAGIC_MASK, 0);
+            device.WriteRegister(DS90UB9x.MAGIC, 0);
             device.WriteRegister(DS90UB9x.MAGIC_WAIT, 0);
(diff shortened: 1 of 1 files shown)


aacuevas (changes requested):
DS90UB9x.MAGIC should be also set to 0 for safety. Apart from that, seems good.

jonnew:
Oops, should be good now. @bparks13 I think this is ready

bparks13 (approved):

Issue #408, opened 2025-02-11
Deadlocking issues


jonnew:
There are situations where a workflow appears to deadlock. The lastest is when i ran the attached workflow against library version 0.4.3. The workflow "runs" but there is no probe B on my test setup and the NP2.0 calibration file paths were wrong. I was presented with a text box indicating this. I accepted this, the ports were shut down but Bonsai remains in the running state:

![Image](https://github.com/user-attachments/assets/e9baf3d3-0994-453d-a6d5-2e3e93eb795c)

Clicking the red square does nothing. There is a logic issue that we need to solve and might be related to #406.

However, it alsohighlights the need for the following:

https://github.com/bonsai-rx/bonsai/issues/1868

If bonsai is an IDE then it needs a mechanism for force stopping workflows that are misbehaving.

Issue #406, opened 2025-02-05
Deadlock or race condition preventing disposal if headstage never produces frames


jonnew:
I'm dealing with a pretty nasty issue in which the hardware is left in a complete broken state in some circumstances. The issue seems to stem from a situation in which a headstage is locked and matches the configuration (shows up in device table etc), but, for whatever reason, never produces a frame. This leads to a series of failures

- Stopping bonsai: hardware is left in running state
- Closing bonsai: ports remain on and locked

This situation appears to result in infinite loop until complete power cycle of the ONIX hardware indicating some interplay between the context configuration and software locking and disposal sequences. 

If I go into `oni-repl` and manually turn off all ports, make sure that the context is in non-passthrough mode, etc, then i can recover functionality.

Clearly this is non optimal. there should be a catch all at both Start and Stop actions that completely reset ONIX hardware.

jonnew:
To be honest @aacuevas I think this is actually equally on software and hardware. The problem with register values that are unaffected by a reset is that there is no easy way to get the hardware to default state in catastrophic situations. It makes we want a secondary reset signal, `reset_hard` that brings all device registers to power-on default.

jonnew:
This may be fixed by #409, but I'm not confident enough to close it. The issue is quite hard to reproduce.

Issue #401, opened 2025-01-17
Hardware stays engaged for 30 seconds after workflow is stopped


ChucklesOnGitHub:
I've encountered this several times while using a miniscope and while using only a breakout board.
After pressing stop, the visualizations shut down but the breakout board "Running" light continues to blink as during acquisition.
In the case of the miniscope, the blue LED continues to be on.
During this period, pressing Start throws a  "Attempt to trigger an already triggered operation" error.
After a while of waiting without further action (can be 30ish seconds), acquisition finally stops as in the "Running" LED on the breakout stays on and the miniscope blue LED turns off.
See video [here](https://drive.google.com/file/d/1NKaDB7KcPHfIuIEZ2ILDXUukshlWGcxF/view?usp=sharing)

Attached is the offending miniscope workflow 
[test_miniscope.zip](https://github.com/user-attachments/files/18455557/test_miniscope.zip)

but even this simple breakout board one had this issue for me too:
![Image](https://github.com/user-attachments/assets/9c8790d0-f45d-465e-87a9-9cc422bea61d)

The system sometimes says "Failure to write to a stream or register" even though I've changed nothing from the previous run and I have to press Start again for it to work (without changing anything and the "Failure to write to a stream or register"  does not reappear). This is quite frequent, I ignore the popup and try again.

Maybe this is just my system but I can't figure out how to correct it. This is on a fresh software I re-installed today.

ChucklesOnGitHub:
hi @jonnew have you been able to figure out what might be going on here?

emilyjanedennis:
Hey Ceci we also have this issue with our board when we're using just the analog in for TTLs. It can take a few minutes for us

jonnew:
- @emilyjanedennis Your issue may be addressed with #409. However, to know for sure I can try to reproduce, can you share your workflow?
- @ChucklesOnGitHub At least part of your issue is a result of something that is a problem from the users perspective, but actually is not a logical error or but in the code. It occurs because @ChucklesOnGitHub used [this example workflow](https://open-ephys.github.io/bonsai-onix1-docs/articles/hardware/ucla-miniscope-v4/overview.html) and removed the miniscope to test the breakout in isolation.  In this case,  the only device that was producing data was the heartbeat. Further, in that workflow, which must have been produced using a old version of the library, the heartbeat rate is set to to 10 Hz

[...]

PR #409, opened 2025-02-12
Tie lifetime of oni.Context to that of ContextTask

### LINKED ISSUE #407: ContextTask.Reset() should call oni.Context.Refresh() instead of deleting and reopening context
During the initialization process, we are resetting the context multiple times to refresh the device table when conditions change (e.g.: after powering a headstage or setting passthrough mode).

At the moment `ContextTask.Reset()` disposes the entire `oni.Context` and creates a new one. This seem inefficient. The reset should instead call `oni.Context.Refresh()`, which issues the soft-reset command and repopulates the device table without closing the context or hardware connection itself.

This is required if we add a hardware safety that truly resets the hardware to POR conditions after the connection is closed, as suggested on #406 .

@glopesdev can you think, top of your head, of any possible downside that I might not be seeing?

jonnew:
- Instead of disposing and recreating oni.Context when a reset occurs, issue a oni.Context.Refresh() to existing object
- Fixes #407 

## Hardware Tests
I tried to put this throgh the ringer

|Test  |Desc  |Result  |
|---|---|---|
| Breakout only | Stated with breakout only, quickly hitting refresh during acqusition etc.  |  ✔️|
| Breakout , NP2.0e | Added np2.0e headstage, which reliese on passthrough configuration, quickly hitting refresh during acquisition etc. |  ✔️|
| Breakout, NP2.0, Rhs2116 headstage | A Rhs2116 headstage,  which uses a standard port but needs to issue a [second reset](https://github.com/open-ephys/bonsai-onix1/blob/067989b21f38169dad5282ce92b971120fe2d76e/OpenEphys.Onix1/ConfigureHeadstageRhs2116.cs#L148) after power up:  | ✔️ |
| Breakout, NP2.0, Rhs2116 headstage with auto voltage | Same as last but do auto voltage discovery on rhs2116 headstage  | ✔️ |

### DIFF
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -35,7 +35,7 @@ namespace OpenEphys.Onix1
     /// </remarks>
     public class ContextTask : IDisposable
     {
-        oni.Context ctx;
+        readonly oni.Context ctx;
 
         /// <summary>
         /// Maximum amount of frames the reading queue will hold. If the queue fills or the read thread is not
@@ -115,11 +115,8 @@ namespace OpenEphys.Onix1
                     lock (readLock)
                         lock (writeLock)
                         {
-                            if (ctx != null)
-                            {
-                                ctx.Refresh();
-                                Initialize();
-                            }
+                            ctx.Refresh();
+                            Initialize();
                         }
                 }
         }
@@ -289,8 +286,9 @@ namespace OpenEphys.Onix1
                     }
 
                     if (!acquisition.IsCompleted)
+                    {
                         throw new InvalidOperationException("Acquisition already running in the current context.");
-
+                    }
 
                     // NB: Configure context before starting acquisition or the the settings (e.g. Block read
                     // and write sizes) will not be respected
@@ -486,7 +484,9 @@ namespace OpenEphys.Onix1
             lock (disposeLock)
             {
                 if (!disposed)
+                {
                     action();
+                }
             }
         }
 
@@ -562,8 +562,7 @@ namespace OpenEphys.Onix1
                     lock (readLock)
                         lock (writeLock)
                         {
-                            ctx?.Dispose();
-                            ctx = null;
+                            ctx.Dispose();
                         }
         }
 
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -93,12 +93,12 @@ namespace OpenEphys.Onix1
             groupedFrames.Connect();
             contextDriver = driver;
             contextIndex = index;
+            ctx = new oni.Context(contextDriver, contextIndex);
             Initialize();
         }
 
         private void Initialize()
         {
-            ctx = new oni.Context(contextDriver, contextIndex);
             SystemClockHz = ctx.SystemClockHz;
             AcquisitionClockHz = ctx.AcquisitionClockHz;
             MaxReadFrameSize = ctx.MaxReadFrameSize;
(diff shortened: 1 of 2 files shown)


aacuevas (approved):
Seems to be working with no apparent downside indeed.

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -93,12 +93,12 @@ internal ContextTask(string driver, int index)
             groupedFrames.Connect();
             contextDriver = driver;
             contextIndex = index;
+            ctx = new oni.Context(contextDriver, contextIndex);
```
glopesdev, line 96: Looks good. My only comment is since the context is no longer disposed and replaced anywhere else other than through the `Dispose` method, and that it is initialized in the constructor, we could consider making this variable `readonly`.

This would mean we never have to check it for null and would place useful constraints on accidental future changes

jonnew: Thanks @glopesdev sounds good!

jonnew:
Looks like there might be an issue with Miniscope documented by @ChucklesOnGitHub 

image

the red lines are indications of issues #406, #401, #408. So these changes _might_ fix that, but the miniscope is now having issues. I will try to reproduce this.

jonnew:
@ChucklesOnGitHub I am unable to reproduce this issue. This is the workflow im using: 

image

Note that miniscope voltage autodiscovery is being used. No issues running this and capturing analog data from breakout and image data from miniscope. 

Can you tell us exactly what error you're encountering? You can right click and copy the message at the bottom left of the bonsai editor when an exception occurs to get a detailed log.

ChucklesOnGitHub:
For me, the issue was present using a workflow with only the breakout board: 
image

Only in test 2 did the issue appear when using the miniscope, all other miniscope tests (4,6,8 in the table above) were fine.
image

jonnew:
Hi @ChucklesOnGitHub, I need clarification about the word "issue". 

From the table, I think "issue" refers to the large issue that this PR is attempting to solve. Namely, that in some cases, after stopping the bonsai workflow, the hardware remains engaged as if it was acquiring data instead of stopping normally. Lets call this issue 1.  The second issue that I see, is that with this PR applied, there is another secondary issue that appears which is a "Failure to write stream or register" exception. This is a problem, but one that is actually in line with  normal operation and some communication issue happening. Let's call this issue 2. 

Is this assessment correct? If so, I am not able to reproduce issue 2. In order to figure out if its being created as a result of this PR, i need to know where its coming from. What is the Node in the workflow that produces the exception and if you can copy to the extended text of the exception that would be helpful.

jonnew:
OK, so @ChucklesOnGitHub and I met out of context to clarify the situation. It turns out that there a couple of separate issues. 

- First of all, Issue 2 in above definition appears to be read herring. Upon revisiting PR code, the miniscope is OK, or at least as OK as the auto-voltage discovery will allow given its inherent limitations
- Issue 1 is also solved with this PR
- A final issue (#401) is that even when stopping the workflow did shutdown the hardware, sometimes this took up to 30 seconds after the workflow was stopped to occur. This was a result of something that is a problem from the users perspective, but actually is not a logical error or but in the code. It occurs because @ChucklesOnGitHub downloaded this example workflow and removed the miniscope to test the breakout in isolation.  In this case,  the only device that was producing data was the heartbeat. Further, in that workflow, which must have been produced using a old version of the library, the heartbeat rate is set to to 10 Hz

  image

  It should be hardcoded to 100 (#359). Combined with the large `ReadSize` that was originally required for the Miniscope, the read buffer took very long to fill in order to allow the `read_frame()` call to return and hardware shut down to proceed. I'm going to address this separately in #401.

PR #410, opened 2025-02-18
Update issue templates

jonnew:
- Tried to provide instructions on how to "attach a workflow"
- I have a nice gif of this process, but I don't think i can put it in the template


bparks13 (approved):
Approved. With these changes users who try to submit a bug report will be required to describe the bug, provide steps on how to recreate it, choose the Bonsai and OpenEphys.Onix1 library versions, and upload the XML of their workflow before they can submit it. We still have the ability to choose a blank issue if we need to, for development purposes

PR #411, opened 2025-02-18
Update bug report template to use customizable web form fields

bparks13:
## How the form is presented to users

image

## How the form shows up as an issue

image


jonnew:
Wow this is really cool. Overall very nice.

- I would remove the title [Bug]: thing. We can classify this later using tags, just let the user write something.
- I was wrong about the library version being deducible from the workflow. Can we add something that forces the user to select a version or just write one. If the later, then instructions on how to locate that should be provided.

PR #417, opened 2025-02-25
Improve error reporting for Neuropixels 1.0e/f

### LINKED ISSUE #370: Improve error reporting for Neuropixels V1f
It would be helpful to know which probe is not seated properly, if possible.

image

bparks13:
- When attempting to read the probe number, if there is an error report the address of the probe that is failing

- Fixed #370


jonnew (approved):
Looks good.

PR #418, opened 2025-02-25
Modify Neuropixels aggregate names

### LINKED ISSUE #371: Name all ConfigureHeadstage nodes the same
This is probably unattainable, but here goes. It gets confusing to sort through headstages and devices but this could be easier if the neuropixels headstages configure nodes were called ConfigureHeadstageNeuropixelsXXX so all the nodes to configure headstages appear listed together.

image

 Or maybe hide the ConfigureDevice nodes in a library for developers? Just thought I'd ask :)

bparks13:
To make sorting easier, all aggregates now follow the pattern of ConfigureHeadstage*.
To aid in the transition, the original names are kept and inherit from the new name, but with an Obsolete tag indicating they will be removed in the future.

Miniscope was not included in this PR, as it is something of a special case where the aggregate and the device are very closely tied together.

- Fixes #371

### DIFF
--- a/OpenEphys.Onix1.Design/NeuropixelsV1eHeadstageDialog.cs
+++ b/OpenEphys.Onix1.Design/NeuropixelsV1eHeadstageDialog.cs
@@ -3,10 +3,10 @@
 namespace OpenEphys.Onix1.Design
 {
     /// <summary>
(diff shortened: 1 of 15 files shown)


jonnew (approved):
I committed some changes, please proofread before merging.

bparks13:
Proofread, everything looks good.

PR #420, opened 2025-02-26
Update Miniscope properties

### LINKED ISSUE #374: Unify Miniscope properties with Open Ephys.Miniscope library?
Not sure if this is desirable, but might be helpful for users migrating from one system to another 

- electrowetting lens should have different range if the intention is to match the Miniscope DAQ library.

- The values in the drop-down of the FPS say both FPS and Hz, might not have been intentional

bparks13:
Update the Focal parameter to match `OpenEphys.Miniscope` as seen [here](https://github.com/open-ephys/bonsai-miniscope/blob/4eddb5c8046159b9e40d689450f7f2b75a3aeac8/OpenEphys.Miniscope/UclaMiniscopeV4.cs#L38).

Removed the `Hz` suffix from the frame rate variable, while maintaining backwards compatibility.

- Fixes #374


jonnew (approved):
I've made some edits directly. Please proofread and test before merging.

PR #434, opened 2025-04-18
Add option to invert neuropixels data

### LINKED ISSUE #419: "Inverted" data from neuropixelsV<X>e headstages
The data produced by both neuropixels 1.0e and neuropixels 2.0e headstages is "inverted". Its centered around 2^9 and 2^11, respectively. But the voltages that are recorded seem like they are produced by electrodes tied to the inverting input of an input amplifier:

Image

I'm trying to think if there is anyway this could be happening due to our use of the passthrough device to get these data?

jonnew:
- For 1f this is done in hardware by manipulating the fixed point gain register values
- For the others, its done in software
- Needs GUI support
- Fixes #419

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <PackageIcon>icon.png</PackageIcon>
-    <VersionPrefix>0.4.5</VersionPrefix>
+    <VersionPrefix>0.5.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <LangVersion>10.0</LangVersion>
     <Features>strict</Features>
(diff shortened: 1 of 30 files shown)


jonnew:
Needs to be tested in hardware.

bparks13 (changes requested):
I've added the UI elements so that the user can change the invert polarity option from the GUIs. Tested saving the values in the GUI, and also cancelling the dialog, to ensure that all user inputs are respected.

Upon testing with both 1.0f and 2.0e headstages, I'm not certain that the inversion is implemented correctly. 1.0f might be working as expected, but with the probe suspended in air it is hard for me to tell. 2.0e is definitely not working as it is overflowing the ushort so that data is wrapping from the maximum value (65535) back to zero, causing huge jumps in the data. I can work on this later if needed, but i will leave it here for now.

jonnew:
@bparks13 Although I agree that this inversion should be applied to both probes and there  should be no per-probe option ,the placement of the checkbox makes the user expect otherwise:

image

The highlighted option is the only one that does not apply per probe. Why is it here?

PR #436, opened 2025-04-22
Display automatically set port voltage value

### LINKED ISSUE #429: Feature request: inform users the result of the autovoltage detection
This could be a grayed out property where the voltage override is.
Helpful for troubleshooting with users when there are headstage communication issues.

jonnew:
- When the PortVoltage is null, the port voltage will be set automatically. If this procedure succeeds, the voltage will be displayed in the text box as "X.X (Auto)" where X.X is the voltage that is applied to the port.
- This text will only appear while the workflow is running. It will revert back to null after the workflow is stopped.
- If the voltage is set manually, the value will persist.
- Fixed #429

### DIFF
--- a/OpenEphys.Onix1/ConfigurePortController.cs
+++ b/OpenEphys.Onix1/ConfigurePortController.cs
@@ -21,7 +21,8 @@ namespace OpenEphys.Onix1
                      "voltage discovery. Warning: Supplying excessive voltage may result in damage to devices. " +
                      "Consult the device datasheet and documentation for allowable voltage ranges.")]
         [Category(ConfigurationCategory)]
-        public double? PortVoltage { get; set; } = null;
+        [TypeConverter(typeof(PortVoltageConverter))]
+        public AutoPortVoltage PortVoltage { get; set; } = null;
 
         protected virtual bool CheckLinkState(DeviceContext device)
         {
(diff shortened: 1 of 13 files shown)


bparks13:
@jonnew I've left the applied voltage at whatever value it was set to so that the value found is persistent even after the workflow closes. I also fixed an exception that would be thrown if the node is first placed and the entire member variable was null.

I saw that you specifically set the applied voltage to be null after the workflow ends, but after testing I believe that we can leave it written in the property grid without issue. The reason is because the `Requested` amplitude is still null since the user at this point has not chosen a voltage override. This means that the automated algorithm will still run even if there is a value displayed in the property. If the user then decides to write in a new value, they will need to erase the `(Auto)` portion of the string and replace the value with whatever they want, but this behavior was true before the changes made so I don't think it's a big change.

jonnew:
Looks good. I kept that and added a custom exception message too. Tell me what you think.

OpenEphys.Onix1/ConfigurePortController.cs
```diff
@@ -77,13 +77,18 @@ public override IObservable<ContextTask> Process(IObservable<ContextTask> source
...
                 if (!serdesLock)
                 {
-                    dispose();
                     var port = (PortName)deviceAddress;
                     var portString = port.GetType()
                                          .GetField(port.ToString())?
                                          .GetCustomAttribute<DescriptionAttribute>()?
                                          .Description ?? "Address " + deviceAddress.ToString();
-                    throw new InvalidOperationException($"Unable to acquire communication lock on {portString}.");
+                    var appliedVoltage = PortVoltage.Applied;
```
bparks13, line 85: Why is this value pulled out but never used?

bparks13 (approved):
Looks good to me, I only have one comment to address before merging

OpenEphys.Onix1/ConfigurePortController.cs
jonnew: good catch

Issue #384, opened 2024-12-12
Hardware timestamp when a stimulation is triggered and save stimulation waveform parameters


cjsha:
It is probably important for users that are utilizing stimulation capabilities to know when a stimulus is triggered. We can do this in software, but we should probably use a hardware timestamp here.

Users probably also need a way to know the shape of the stimulation waveform each time a stimulus is triggered. I think the best way is to output the stimulation waveform parameters and save them to a .csv every time stimulation is triggered (with the hardware timestamp?)

jonnew:
What operators and hardware does this refer to?

cjsha:
Relevant hardware:
- HS64
- RHS2116

Relevant operators:
- Rhs216StimulusTrigger
- Headstage64OpticalStimulatorTrigger
- Headstage64ElectricalStimulatorTrigger

jonnew:
This was a big oversight on my part when making the gateware for these devices. I'm in the middle of update each of them to provide a frame indicating stimulus times and states for easy synchronization with other data streams. We will need to revisit this to implement software support once that is done.

PR #442, opened 2025-04-30
Add support for stimulus trigger events from headstage-rhs2116

jonnew:
- New trigger event data operators
- Add recovery mask to existing rhs2116 data operators
- Create new, explicitly Rhs2116PairDataFrame type because it was going to be a hack to reuse Rhs2116DataFrame with this new recovery information
- Partially addresses #384


bparks13 (approved):
This looks good from a code review standpoint, but I have no hardware to test this for functionality. If it has been tested in hardware then this is good to merge.

PR #448, opened 2025-05-20
Make neuropixel channel maps fixed size arrays

### LINKED ISSUE #280: ChannelMap should be a fixed size array
In the case of all neuropixels yet supported, this should hold 384 elements:

https://github.com/open-ephys/onix-bonsai-onix1/blob/3fdcec67f5f5d548ddf9125d7859565b55a48d08/OpenEphys.Onix1/NeuropixelsV1eProbeConfiguration.cs#L156

Therefore, `List` is probably not the right type. If we change to an `array`, we must diligently propagate the array interface downstream (we cant just perform `ToList()` at the next step).

jonnew:
- They were Lists but are required to have a fixed number of elements
- Various other simplifications related to this change
- Fixed #280

### DIFF
--- a/OpenEphys.Onix1.Design/NeuropixelsV1ChannelConfigurationDialog.cs
+++ b/OpenEphys.Onix1.Design/NeuropixelsV1ChannelConfigurationDialog.cs
@@ -247,7 +247,7 @@ namespace OpenEphys.Onix1.Design
             return index.ToString();
         }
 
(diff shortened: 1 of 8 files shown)


bparks13 (approved):
Looks good, tested this with Neuropixels 1.0f and 2.0e headstages, the channel maps are intact and can be modified using the GUIs without issue. Writing the configuration also worked without error, and data was streaming as usual.

PR #450, opened 2025-05-23
Add 1 second delay before applying port voltage to headstage-64

### LINKED ISSUE #430: headstage lock check sometimes fails spureously
Sometimes, in a hard to consistently reproduce way, headstage link detection fails with the "Unable to obtain lock" message.
When this happens:
- It is immediate (i.e.: no actual port sweep seems to be taking place)
- It might happen even if a fixed voltage is set

So it might be some _other_ actual exception that just occurs on the startup process and this block, in turn, throws the unable to obtain lock one, or some locking issue.

A couple more details:
- When this happens, pressing Bonsai's restart workflow button seems to trigger this issue more reliably, which suggest some cleaning up issue or timing-related issue
- We have observed this on HS64, not sure yet if this happens with other headstages as well
- Workflow topology (how many operators, visualizers, etc...) seems to affect the probability of this happening, which also points to a deferred resource disposal clash or similar issue.

jonnew:
- During a restart, the headstage is turned off and then on again very rapidly. This seems to cause the serializer not to be completely shut down and to be in an indeterminate state. To combat this, I've simply added a delay before the first voltage to turn the headstage on is applied.
- Fixes #430

### DIFF
--- a/OpenEphys.Onix1/ConfigureHeadstage64.cs
+++ b/OpenEphys.Onix1/ConfigureHeadstage64.cs
@@ -148,6 +148,17 @@ namespace OpenEphys.Onix1
 
         class ConfigureHeadstage64PortController : ConfigurePortController
         {
+
+            protected override bool ConfigurePortVoltageOverride(DeviceContext device, double voltage)
+            {
+                // NB: Wait for 1 second to discharge the headstage in the case that they have e.g. just
+                // restarted the workflow automatically with nearly no delay from the last run. 
+                Thread.Sleep(1000);
+                device.WriteRegister(PortController.PORTVOLTAGE, (uint)(voltage * 10));
+                Thread.Sleep(500);
+                return CheckLinkState(device);
+            }
+
             protected override bool ConfigurePortVoltage(DeviceContext device, out double voltage)
             {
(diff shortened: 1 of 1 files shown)


jonnew:
I should also note that we might want to consider doing something similar to the other headstages (although they don't, in general, seem to have nearly as much difficulty as hs-64).

jonnew:
Interesting. For the first issue, I think we are now at the hardware limitations. Basically, it seems to making it through the discovery procedure, which requires a valid lock, but then then in between this and actually talking to the device, the lock is lost. This is pretty much a hardware limitations. 

The second issue "Attempt to trigger.." is entirely different and warrants opening a new issue using this report and video. Ill do that.

ChucklesOnGitHub:
It's the devices in the example workflow: 
image

jonnew:
@ChucklesOnGitHub  I have reliably reproduced this issue. Its a secondary one that we will need to solve. I can see why its happening, but I don't know why its not handled. I will create a new issue documenting this. Good find!!

jonnew:
@ChucklesOnGitHub Here's the issue: https://github.com/open-ephys/bonsai-onix1/issues/455

PR #451, opened 2025-05-23
Change DigitalPortState backing type from ushort to byte

### LINKED ISSUE #449: Consider saving DigitalPortState as byte instead of ushort
In preparation for saving and loading digital inputs in Bonsai 2.9 (e.g. using the unmanagedtype overload for MatrixWriter), I came across this:

https://github.com/open-ephys/bonsai-onix1/blob/c57cc7c2b726122777f8ced3999cb086ae5f01dd/OpenEphys.Onix1/ConfigureDigitalIO.cs#L87

Since there are 8 pins, we can fit all digital pin data in a byte instead of a ushort which would cut the storage required for digital inputs in half. However, there might be a reason for not doing this that I don't know or understand yet.

jonnew:
- MSBs are set to 0 by hardware so why waste the space
- Fixes #449

### DIFF
--- a/OpenEphys.Onix1/ConfigureDigitalIO.cs
+++ b/OpenEphys.Onix1/ConfigureDigitalIO.cs
@@ -84,7 +84,7 @@ namespace OpenEphys.Onix1
     /// Specifies the state of the ONIX breakout board's digital input pins.
     /// </summary>
     [Flags]
-    public enum DigitalPortState : ushort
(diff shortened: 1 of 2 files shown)


aacuevas (approved):

aacuevas:
The endianess comment is a good point, we need to review datasheets that use byte types to check that the data structure in them matches the actual data sent

PR #452, opened 2025-05-26
Allow single PolledBno055Data operator per device name

### LINKED ISSUE #316: Polling Data operators should reference a single underlying polling loop
For instance, lets say the user puts in two `PolledBno055Data` nodes, each with the same `DeviceName` arguments. These two nodes will now share access to a single underlying i2c bus instead of just pulling data from a common, underlying polling loop. This makes them appear to have half the sample rate each since the bandwidth is limited by i2c communication speed.

jonnew:
- Use static HashSet to hold names of DeviceNames that have been used to create a Subscription to upstream sequence.
- Explicitly remove DeviceNames from HashSet by modifying the IDisposable returned by Subscribe that is called at sequence end.
- Fixes #316


aacuevas (approved):
If this were a more general issue we might want to look for a different approach, but since this is an issue very localized to this operator, and since we want to avoid having to implement anything polled in the future, I believe this works nicely.

bparks13 (approved):

PR #454, opened 2025-05-29
Document Rhs2116Trigger's input double as delay

### LINKED ISSUE #453: Document Rhs2116Trigger's input double as delay

cjsha:
Doubles received by the Rhs2116Trigger node indicate the delay in microseconds before delivering the stimulation waveform.

Fix #453


jonnew (changes requested):
Although this is needed, we actually need to specify that its both a delay and a trigger. To be clear what this does is send a value to the headstage. As soon as that value is received, the headstage starts an internal timer that counts out the value number of milliseconds and then delivers the programmed stimulus.

jonnew (approved):

PR #456, opened 2025-05-31
Add support for i2c multibyte access and make polledbno use it

aacuevas:
Capitalizes on the gateware changes for multibyte i2c (required for this to work)

Measured a significant performance improvement, which brings bno polling to around 90Hz with all fields enabled.

### DIFF
--- a/.bonsai/Bonsai.config
+++ b/.bonsai/Bonsai.config
@@ -2,21 +2,29 @@
 <PackageConfiguration xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <Packages>
     <Package id="Bonsai" version="2.8.5" />
+    <Package id="Bonsai.Arduino" version="2.8.1" />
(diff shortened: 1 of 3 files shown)


bparks13:
@jonnew @aacuevas Do we want this PR to be a part of 0.5.0? I don't see a milestone attached to it, not sure if it's worth trying to fit in here or not

aacuevas:
No, this requires a gateware change to work, so it needs to be on 0.6, which will be released alongside gateware version 2.0. We need to be very careful with that.

jonnew (changes requested):
See comments in the code and PR.

jonnew:
For the curious, here is what happens when we capture only Temperature and drive the node with a 1 kHz heartbeat:

image

Text values are in milliseconds.

aacuevas:
Updated according to all comments

jonnew (approved):
I've done a bit of simplification to the polling logic and bumped the version. This is ready to merge.

jonnew:
I'm going to rebase this using persistent-heartbeat and retarget the PR since this a 0.6.0 target.

PR #464, opened 2025-06-24
Add version check for onix gateware v2.x

### LINKED ISSUE #447: Firmware version check
Starting with version 0.6.0 there should be a firmware version check. The reason for this is that, starting in 0.6.0 version 1.1 of the controller firmware will be required, namely the addition of persistent heartbeat to the controller. In general, a hardware version and firmware version check is a good thing to have. 

Some thought should go into making this class and functionality. For instance, is it a good idea to do checks for clroni, liboni, and driver translator versions as well? This is an open question.

aacuevas:
- Fixes #447 

I am not sure if the message is OK as it is. Specially the URL. It was intended to help users, but since it is not clickable, I am not sure if it is appropriate or even helpful at all.

### DIFF
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -76,8 +76,6 @@ namespace OpenEphys.Onix1
         readonly string contextDriver = DefaultDriver;
         readonly int contextIndex = DefaultIndex;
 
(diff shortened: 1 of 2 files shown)


jonnew (approved):
I made some changes. Please have a look at these to make sure I didn't completely screw something up.

PR #472, opened 2025-07-21
Add TaitBryanAngles type

jonnew:
- This is a partial wrapper for Vector3 with all arethmetic operations removed since they largely dont make sense and X, Y, and Z mapped to yaw, pitch, and roll
- Use this type in Bno055DataFrame

### DIFF
--- a/OpenEphys.Onix1.Design/EulerAnglesVisualizer.cs
+++ b/OpenEphys.Onix1.Design/EulerAnglesVisualizer.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using Bonsai;
+using Bonsai.Design.Visualizers;
+using OpenEphys.Onix1.Design;
+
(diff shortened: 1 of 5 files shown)


jonnew:
> almost wonder if this should go in some common library, it can pertain to onix1, acquisition-board, and miniscope packages.

Yes, I thought of this. I guess we could make a Bonsai package that provides this type, the visualizer, and a source node that could be used by both Miniscope and this.

aacuevas (approved):
In any case, for now I believe it's ok.
A dumb question, should we add a copy constructor to the class?

Right now, if we do 
```
TaitBryanAngles a = new TaitBryanAngles(Y,P,R);
TaitBryanAngles b = new TaitBryanAngles(a);
```

It works because of the implicit conversion to Vector3 and then using the Vector3 constructor. I am not sure if we should add the direct copy constructor. Not for performance (there's not gonna be a difference) but for readability.

PR #473, opened 2025-07-22
Update StartPosition to CenterParent

### LINKED ISSUE #311: Set all dialogs to open in the center of their parent
In previous commits, panels were updated to eliminate the percent scaling issue that was being seen; while this is fixed if the dialog is opened in the same screen as the original Bonsai window, it is not correct if Bonsai is on a different window than the dialog is opened in. The fix is easy; set `StartPosition` to `CenterParent` for all dialogs that can be opened directly from Bonsai. This will ensure that the same resolution is used for the new window, and it should not have any issues with scale.

bparks13:
This PR fixes the last two dialogs to open up using the `CenterParent` value, which means that the dialog will open up on top of the parent window that called it. This is useful for a number of reasons, including ensuring that the screen resolution is consistent between windows, and also for ease of use by the user.

- Fixes #311

### DIFF
--- a/OpenEphys.Onix1.Design/ChannelConfigurationDialog.Designer.cs
+++ b/OpenEphys.Onix1.Design/ChannelConfigurationDialog.Designer.cs
@@ -51,7 +51,7 @@
             this.splitContainer1.Dock = System.Windows.Forms.DockStyle.Fill;
             this.splitContainer1.FixedPanel = System.Windows.Forms.FixedPanel.Panel2;
             this.splitContainer1.IsSplitterFixed = true;
-            this.splitContainer1.Location = new System.Drawing.Point(0, 30);
+            this.splitContainer1.Location = new System.Drawing.Point(0, 24);
             this.splitContainer1.Margin = new System.Windows.Forms.Padding(3, 2, 3, 2);
             this.splitContainer1.Name = "splitContainer1";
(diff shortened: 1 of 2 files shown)


jonnew:
I'm unsure how to test this commit. Do you have suggestions?

Or, if you have screenshots of the before and after behavior that would be good.

bparks13:
@jonnew this can be tested if you have monitors with two different resolutions; I did not necessarily see any issues outside of that scenario. I do not have screenshots of the behavior, but if the main Bonsai window is moved from a screen with one resolution to a different screen with a different resolution and then one of the GUIs was opened, it could open the GUI in the first window but with the wrong scaling factors due to the difference in screen resolution. It is not guaranteed to open in the wrong window, as Windows could choose where to place the new window, but in the case above it would make the GUI unusable as the controls would be scaled and shifted such that they were hidden or squeezed to become unreadable.

In summary, to test this commit, you can open Bonsai and move it between different windows, opening and closing GUIs to ensure that the controls are always rendered appropriately.

jonnew:
I tried to test this by setting one of my monitors to a different resolution before and after the commit and I saw no change in behavior. This is likely due to the fact that the monitors dont actual have different physical resolutions.

jonnew (commented):
Can you have a look at my comments.

PR #481, opened 2025-08-11
Move Probe Interface configuration files to external JSON files

### LINKED ISSUE #366: ProbeConfiguration is deserializing ChannelConfiguration to XML
Hi!

Not sure if this is intended, but when adding a `NeuropixelsV2eHeadstage` node to the workflow, a large amount of data (>1Mb) is added to the .bonsai file. 

Is the deserialization of this field really required? Having it this way, makes it extremely difficult to maintain code in GitHub due to its sheer size but also when it comes to go over diff checks.

For reference I am attaching the `bonsai.config` to reproduce the environment and a simple snippet that reproduces the behavior i am looking at.

[ExampleHeadstageOnix.zip](https://github.com/user-attachments/files/17626530/ExampleHeadstageOnix.zip)

bparks13:
## Overview
This PR refactors how probe configuration data is stored and managed across `Neuropixels 1.0`, `Neuropixels 2.0`, and `RHS2116` devices. The main changes involve externalizing probe interface files into JSON format and improving error handling around file operations.

## Key Changes
- Moved calibration file paths into respective `ProbeConfiguration` classes for better encapsulation
- Added `ProbeInterfaceFile` properties to manage external JSON configuration files
- Improved error handling for file operations with specific exception types
- Renamed UI elements from "Channel Configuration" to "Probe Group" for consistency
- Added `GenericPropertyConverter` for better property management in UI
- Cleaned up unused dependencies and imports

## Breaking Changes
1. Calibration file properties have been moved:
   - From `ConfigureNeuropixelsV1`/`ConfigureNeuropixelsV2` classes to their respective `ProbeConfiguration` classes
   - Old properties are marked as obsolete and will be removed in v1.0.0

2. Configuration files are now stored externally:
   - JSON files with pi_*.json extension for automatically created file paths
   - Generated automatically when saving Bonsai workflows
   - Can be manually specified via `ProbeInterfaceFile` property

## Migration Guide
For existing workflows:
1. Calibration files will be automatically migrated to the new ProbeConfiguration location
    - This is only true for `Neuropixels 1.0f` and `Neuropixels 2.0e/Beta` probes. Due to how the properties are organized for `Neuropixels 1.0e` they cannot be automatically migrated.
2. Probe interface files will be generated automatically next to .bonsai files
3. No manual changes required for existing configurations

Fixes #366

### DIFF
--- a/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
+++ b/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
@@ -37,7 +37,7 @@ namespace OpenEphys.Onix1
         {
             Enable = configureNeuropixelsV1e.Enable;
             EnableLed = configureNeuropixelsV1e.EnableLed;
-            ProbeInterfaceFile = configureNeuropixelsV1e.ProbeInterfaceFile;
+            ProbeInterfaceFileName = configureNeuropixelsV1e.ProbeInterfaceFileName;
             ProbeConfiguration = new(configureNeuropixelsV1e.ProbeConfiguration);
             DeviceName = configureNeuropixelsV1e.DeviceName;
             DeviceAddress = configureNeuropixelsV1e.DeviceAddress;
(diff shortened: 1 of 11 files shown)


OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
```diff
@@ -154,7 +162,7 @@ public override IObservable<ContextTask> Process(IObservable<ContextTask> source
             {
                 // configure device via the DS90UB9x deserializer device
                 var device = context.GetPassthroughDeviceContext(deviceAddress, typeof(DS90UB9x));
-                device.WriteRegister(DS90UB9x.ENABLE, enable ? 1u : 0);
+                device.WriteRegister(DS90UB9x.ENABLE, enable ? 1u : 0u);
```
jonnew, line 165: good change, but again not for this issue.

OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
```diff
@@ -82,54 +83,61 @@ public ConfigureNeuropixelsV1e(ConfigureNeuropixelsV1e configureNeuropixelsV1e)
         public bool InvertPolarity { get; set; } = true;
 
         /// <summary>
-        /// Gets or sets the path to the gain calibration file.
+        /// Gets or sets the NeuropixelsV1 probe configuration.
         /// </summary>
-        /// <remarks>
```
jonnew, line 87: Where did this comments end up? I assume they were moved somewhere else but not seeing immediately where.

jonnew:
Something fields incoherent about this:

image

1. The ProbeInterfaceFile is external file you can specify to determine an electrode layout. Its a top level member of the `ConfigureNeuropixels1e`
2. The calibration files are external files you can you use to determine parameters for your probe. They are 2nd order properties inside `ProbeConfiguration`
3. When you open the GUI there are two completely different ways to handle these for historical reasons and decisions. Additionally different terminology is used: `ProbeInterfaceFile` versus `Probe Group`.

I think we need to examine a coherent approach here.

With respect to 1 and 2, why is one file a top level property and the other not? It feels like they should all be inside probe configuration?

With respect to 3, the naming conventions need to be unified and the way things are loaded. Before we had all this import stuff because we had our own internal representation of probe interface inside bonsai file. Now we are just straight up using probe interface files so there is no importing etc as far as I can see.

PR #485, opened 2025-08-21
Reorder tabs to flow from left to right, top to bottom

### LINKED ISSUE #290: Tab order of controls in dialogs is not intuitive
In some of the dialogs (specifically `NeuropixelsV1eDialog`, but probably most other dialogs), when hitting <kbd>tab</kbd> the focus does not just to the next logical control; it will focus on a control that is far away from the previously focused control. This is only noticeable if someone is navigating via the keyboard, but it makes the GUI far less functional from an end-user perspective if the controls are not in a logical tab order.

bparks13:
To make it easier to follow the flow of logic, when pressing <kbd>tab</kbd>, it now follows a logical flow from left-to-right, and top-to-bottom. This is true for all existing dialogs, and this pattern should be followed for all future dialogs added.

**Rhs2116**
example_rhs2116

**Neuropixels V1**
example_npxv1

**Neuropixels V2**
example_npxv2

Fixes #290

### DIFF
--- a/OpenEphys.Onix1.Design/ChannelConfigurationDialog.Designer.cs
+++ b/OpenEphys.Onix1.Design/ChannelConfigurationDialog.Designer.cs
@@ -67,6 +67,7 @@
             this.splitContainer1.Size = new System.Drawing.Size(609, 550);
             this.splitContainer1.SplitterDistance = 513;
             this.splitContainer1.TabIndex = 0;
+            this.splitContainer1.TabStop = false;
             // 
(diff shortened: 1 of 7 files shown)


bparks13:
`Rhs2116StimulusSequenceDialog` now correctly loops from the last control to the first one. Due to WinForms default tabbing behavior, this form needs a tab handler to close the loop, otherwise the focus could go to the `ChannelConfigurationDialog` and get stuck there.

example_rhs2116_looping

PR #486, opened 2025-08-21
Add tooltip to Invert Polarity button

### LINKED ISSUE #478: Add tooltip to "Invert Polarity" checkbox
I think it's potentially unclear why someone would want to invert polarity even though it's the recommended setting (and thus why the checkbox is checked by default when the GUI is opened). I think this because I'm not sure it's common knowledge that the raw Neuropixels data from the hardware is inverted. I'm pretty sure other software don't even expose this option to users so they might not be accustomed to seeing it.

Image

Of course, this checkbox is explained [here](https://open-ephys.github.io/bonsai-onix1-docs/articles/hardware/np1e/gui.html#inverting-the-polarity-of-ephys-data) in our docs, but I don't think the user should necessarily have to navigate to the website to figure out what's going on in this case. Therefore, I propose adding a tooltip that says something similar to the explanation in our docs.

I will initially place this in 0.6.1, but this issue may be dropped or moved per discussions in our issue-sorting meetings.

bparks13:
Adds a tooltip the Invert Polarity button for both Neuropixels V1 and V2.

image

Fixes #478

### DIFF
--- a/OpenEphys.Onix1.Design/NeuropixelsV1ProbeConfigurationDialog.Designer.cs
+++ b/OpenEphys.Onix1.Design/NeuropixelsV1ProbeConfigurationDialog.Designer.cs
@@ -285,7 +285,7 @@
             this.tableLayoutPanel1.Controls.Add(this.flowLayoutPanel1, 0, 1);
             this.tableLayoutPanel1.Dock = System.Windows.Forms.DockStyle.Fill;
             this.tableLayoutPanel1.Location = new System.Drawing.Point(0, 24);
-            this.tableLayoutPanel1.Margin = new System.Windows.Forms.Padding(4, 4, 4, 4);
+            this.tableLayoutPanel1.Margin = new System.Windows.Forms.Padding(4);
             this.tableLayoutPanel1.Name = "tableLayoutPanel1";
             this.tableLayoutPanel1.RowCount = 2;
             this.tableLayoutPanel1.RowStyles.Add(new System.Windows.Forms.RowStyle(System.Windows.Forms.SizeType.Percent, 100F));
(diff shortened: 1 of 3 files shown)


bparks13:
@jonnew that's a good point. In the interest of DRY, it might be better if we can bind the existing description to the tooltip, so there is only one source of information. Instead of manually setting the tooltip, we could do the following during construction of the GUI:

```C#
var property = typeof(IConfigureNeuropixelsV2).GetProperty(nameof(IConfigureNeuropixelsV2.InvertPolarity));
var attribute = property?.GetCustomAttribute<DescriptionAttribute>();
toolTip.SetToolTip(checkBoxInvertPolarity, attribute != null ? attribute.Description : "");
```

This works, but requires some additional work as described below, and the Description attribute has less information than the XML comment, so all the user would see in the tooltip is: "Invert the polarity of the electrode voltages acquired by the probe." This is all that the user sees from the Bonsai editor already, so it might be sufficient, but it does not give much reasoning on _why_ they might want to invert polarity.

For the reflection above to work, I had to add the Description attribute to the interface, because currently the Description is only present in the inheriting classes. The GUI takes an `IConfigureNeuropixelsV2`, so it doesn't care which device is actually being given; it seemed less confusing and more consistent to move the attribute to the interface than to add code inside the dialog to determine which device is being given.

Looking into this with a critical eye also highlights another area where we are repeating information. The two configuration classes `ConfigureNeuropixelsV2e` and `ConfigureNeuropixelsV2eBeta` both inherit from `IConfigureNeuropixelsV2`, but they independently add XML comments, descriptions, and categories to the inherited elements. Instead, we could consolidate the repeated text into the interface, and only add information when it differs from the interface (i.e., the `Enable` property XML comments use the correct class (`V2e` or `V2eBeta`) when discussing the Data node, this should remain outside of the interface). 

Repeating information is also present in `NeuropixelsV1` classes, where there is an interface with minimal documentation, while the `V1e` and `V1f` classes duplicate information, with some slight differences (e.g., `V1f.Enable` says "Neuro data stream", while `V1e.Enable` says "Neuropixels data stream").

Let me know if this makes sense to or if it is worth putting effort into. For now, we can merge this PR by updating the string manually as you suggested and we can throw this information into a separate issue to be handled later.

PR #491, opened 2025-08-27
Add Ground as reference option for Neuropixels 2.0 probes

### LINKED ISSUE #294: Add GND as reference option for Neuropixels 2.0
Currently the only reference options available for Neuropixels 2.0 probes are `EXT`, and `TIP1-4`. On commercial probes, GND is a valid option for references.

This would only be an option for non-beta probes. We need to go ensure that the correct bits are being turned on/off in the shank config for GND to properly be activated. 

This will only be reflected at the GUI level by an additional entry in the drop-down menu, and all configuration will still be handled by the configuration methods in the base library.

bparks13:
Adds `Ground` as an option for the reference. This is added as part of the reference enum, and is propagated to the GUI drop-down menu.

Fixes #294

### DIFF
--- a/OpenEphys.Onix1/ConfigureNeuropixelsV2eBeta.cs
+++ b/OpenEphys.Onix1/ConfigureNeuropixelsV2eBeta.cs
@@ -200,6 +200,12 @@ namespace OpenEphys.Onix1
                 // configure probe A streaming
                 if (probeAMetadata.ProbeSerialNumber != null)
                 {
+                    if (ProbeConfigurationA.Reference == NeuropixelsV2QuadShankReference.Ground)
+                    {
+                        throw new InvalidOperationException($"Neuropixels 2.0-Beta probes do not provide a Ground reference selection. Please select a different reference" +
+                            $" for {NeuropixelsV2Probe.ProbeA}.");
+                    }
+
                     var gainCorrection = NeuropixelsV2Helper.TryParseGainCalibrationFile(GainCalibrationFileA);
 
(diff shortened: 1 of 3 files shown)


OpenEphys.Onix1/ConfigureNeuropixelsV2eBeta.cs
```diff
@@ -221,8 +227,14 @@ public override IObservable<ContextTask> Process(IObservable<ContextTask> source
                 }
 
                 // configure probe B streaming
-                if (probeAMetadata.ProbeSerialNumber != null)
```
jonnew, line 224: Good catch

OpenEphys.Onix1/ConfigureNeuropixelsV2eBeta.cs
```diff
@@ -221,8 +227,14 @@ public override IObservable<ContextTask> Process(IObservable<ContextTask> source
...
+                if (probeBMetadata.ProbeSerialNumber != null)
                 {
+                    if (ProbeConfigurationB.Reference == NeuropixelsV2QuadShankReference.Ground)
+                    {
+                        throw new InvalidOperationException($"Beta probes cannot have a Ground reference selected. Please select a different reference" +
```
jonnew, line 234: For the first sentence:

Neuropixels 2.0-Beta probes do not provide a Ground reference selection.

jonnew (approved):
Looks good, please have a look at comments

PR #494, opened 2025-09-10
Move Neuropixels XML comments to common interface

### LINKED ISSUE #488: Duplicated XML docs
First referenced [here](https://github.com/open-ephys/bonsai-onix1/pull/486#issuecomment-3214800661):

Neuropixels devices copy a lot of documentation between similar devices (e.g., `NeuropixelsV1e` and `NeuropixelsV1f`) despite the fact they have a common interface `IConfigureNeuropixelsV1`. Move all documentation from the classes to the interface, and also copy the attributes to the interface for the purpose of grabbing the Description attribute for tooltips (see #486).

This is also true for `IConfigureNeuropixelsV2`.

bparks13:
Instead of duplicating XML comments across devices, especially when these devices share a common interface, this PR moves all XML comments to the interface and the implementing classes now inherit the docs. 

We still need to add specific comments for properties that are not in the interface, or when a property needs to reference an external class that is different for each device (e.g., referencing the correct `*Data` node).

Also, consolidated some `Description` attributes so they are consistent.

Fixes #488

### DIFF
--- a/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
+++ b/OpenEphys.Onix1/ConfigureNeuropixelsV1e.cs
@@ -43,15 +43,13 @@ namespace OpenEphys.Onix1
             InvertPolarity = configureNeuropixelsV1e.InvertPolarity;
         }
 
-        /// <summary>
-        /// Gets or sets the device enable state.
-        /// </summary>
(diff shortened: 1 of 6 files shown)


jonnew (changes requested):
Looks good, please have a look at my comments

PR #499, opened 2025-09-16
Fix scale on right side of pane for channel configuration dialogs

### LINKED ISSUE #339: Ruler in neuropixels GUI should be afixed to right Y axis
I would like the ruler to be on the y-axis (could even be made from the y-axis with clever label customization or scaling of neuropixels image being plotted) so that when I pan around the ruler is never out of view:

image

vs:

image

If the ruler was mad from ticks on the y-axis, it will still be visible.

bparks13:
This PR updates how scales are rendered and manipulated in channel configuration dialogs. Previously, they were drawn as objects in the same `GraphPane` as the `ProbeInterface` objects. Now, there is a secondary pane that is created to the right of the main pane that overrides the default Y-Axis labels and tics to simplify the logic required to maintain these scales over time.

The scale pane and the main pane are tied together so that they pan/zoom whenever the user pans/zooms in the main pane. Due to the separation in panes, all panning and zooming are disabled in the scale pane to avoid any issues.

Devices that utilize this new feature:
- `NeuropixelsV1e`
- `NeuropixelsV1f`
- `NeuropixelsV2e`
- `NeuropixelsV2eBeta`

For devices that do not have a scale, such as the `Rhs2116`, the original behavior is retained with a single pane to draw the probe object. 

Fixes #339 

**NeuropixelsV1e GIF**

npxv1e-scale

**NeuropixelsV2e GIF**

npxv2e-scale

**Rhs2116 GIF**

rhs2116-probe

### DIFF
--- a/OpenEphys.Onix1.Design/ChannelConfigurationDialog.cs
+++ b/OpenEphys.Onix1.Design/ChannelConfigurationDialog.cs
@@ -1,11 +1,11 @@
-﻿using System.Drawing;
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Windows.Forms;
-using ZedGraph;
-using System;
 using OpenEphys.ProbeInterface.NET;
-using System.Collections.Generic;
+using ZedGraph;
 
 namespace OpenEphys.Onix1.Design
 {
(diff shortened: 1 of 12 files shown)


bparks13:
During testing, this update solves issue #313 as well. Now, the ChannelConfiguration dialog does not reset the zoom status when resizing.

jonnew (commented):
Almost, please see comment. Also because there will no longer be pan boundaries, the scale bar can keep going into the negative, it should display correctly for the entire area it exists over:

Image

jonnew (changes requested):
See comments.

PR #518, opened 2025-10-07
Update ElectricalStimulator dialog to match OpticalStimulator behavior

bparks13:
`OpticalStimulator` dialog was updated in 2c1c9eb67677b1535408a5525f3b71ebae7e03a0 to disable certain text boxes when the number of pulses/bursts was 1, since a period at that point is undefined.

However, the `ElectricalStimulator` was not updated at the same time. That has been corrected in this PR.

This PR also consolidates the estim/ostim dialogs so that the checks occur at the same place; in the Options dialog, at the smallest scope possible to ensure that the behavior is consistent no matter how the text boxes are updated (i.e., during construction, or during regular use).


jonnew (approved):
Thanks, looks good.

PR #521, opened 2025-10-10
Attempt at combined register/GPIO headstage-64 stim triggers

jonnew:
- This is an attempt at automatically using the SERDES GPIO to trigger stimuli on headstage-64 when the delay is 0
- This solution is imperfect. The big issue with it is that if both stimulator triggers are enabled and receiving input, then the although the hardware will ensure the electrical stimulator takes precedence, the optical stimulator trigger will still issue GPIO toggles. This effectively means the OpticalStimulatorTrigger and trigger electrical stimuli.
- If the GPO operator was stand alone, then this would not be surprising, but it is in this case.


jonnew:
This PR was superseded by #529.

PR #523, opened 2025-10-20
Refactor Neuropixels 2.0 probes in preparation for single-shank probes

bparks13:
This PR was initially a part of #493; it has been split out into this PR as well so that we can review the underlying infrastructural changes independently of the addition of the single-shank Neuropixels 2.0 probes. This is prompted by #522, which would benefit from some of these changes but we do not need the single-shank probes for #522 to be implemented. 

Below are the major highlights of this PR:

- Add Neuropixels V2 probe info interfaces and implementations
- Moved all probe-specific logic from the dialog to the info class in the Design library
- Update all methods/classes to handle the abstract class, with easy scalability to add single-shank probes in the future
- Add clone method for NeuropixelsV2eProbeGroup to ensure that dialogs can be exited without affecting settings
- Update how the ChannelMap is accessed and utilized

### DIFF
--- a/OpenEphys.Onix1/ConfigureNeuropixelsV2e.cs
+++ b/OpenEphys.Onix1/ConfigureNeuropixelsV2e.cs
@@ -2,6 +2,7 @@
 using System.ComponentModel;
 using System.Drawing.Design;
 using System.Reactive.Disposables;
+using System.Xml.Serialization;
 using Bonsai;
 
 namespace OpenEphys.Onix1
@@ -56,6 +57,7 @@ namespace OpenEphys.Onix1
         [Category(ConfigurationCategory)]
         [Description("Probe A configuration.")]
         [Editor("OpenEphys.Onix1.Design.NeuropixelsV2eProbeConfigurationEditor, OpenEphys.Onix1.Design", typeof(UITypeEditor))]
+        [XmlElement(nameof(ProbeConfigurationA), typeof(NeuropixelsV2QuadShankProbeConfiguration))]
         public NeuropixelsV2ProbeConfiguration ProbeConfigurationA { get; set; } = new NeuropixelsV2QuadShankProbeConfiguration(NeuropixelsV2Probe.ProbeA, NeuropixelsV2QuadShankReference.External);
 
         /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


OpenEphys.Onix1.Design/NeuropixelsV2eProbeConfigurationDialog.cs
```diff
@@ -142,390 +156,402 @@ private void FormShown(object sender, EventArgs e)
...
 
         private void SelectedChannelPresetChanged(object sender, EventArgs e)
         {
-            var channelPreset = (ChannelPreset)((ComboBox)sender).SelectedItem;
-
-            if (channelPreset != ChannelPreset.None)
+            switch (ProbeConfiguration.ProbeType)
             {
-                SetChannelPreset(channelPreset);
+                case NeuropixelsV2ProbeType.QuadShank:
```
aacuevas: Same as before, separation of responsibilities should make these specifics not be here

OpenEphys.Onix1/NeuropixelsV2eProbeGroup.cs
```diff
@@ -1,267 +1,42 @@
...
-            var shank = index / NeuropixelsV2.ElectrodePerShank;
-            var offset = shankOffsetX + (shankWidthX + shankPitchX) * shank + 11;
-
-            return (index % 2) switch
-            {
-                0 => offset + 8.0f,
-                1 => offset + 40.0f,
-                _ => throw new ArgumentException("Invalid index given.")
-            };
+            return Activator.CreateInstance(typeof(T), Specification, Version, Probes.Select(probe => new Probe(probe)).ToArray()) as T ?? throw new NullReferenceException($"Issue creating new probe group of type {nameof(NeuropixelsV2eProbeGroup)}.");
```
aacuevas: I think I understand what you're trying to do here, but this might be a case of excessive complication.
It might be easier, and way more readable, to just create a virtual (or abstract) base Clone() and have every derived override with a `Clone() { return new WhateverIAm() }`

Maybe I am wrong, but I don't see the advantage of using reflection here. Reflection is a powerful tool, but should not be overused.

PR #530, opened 2025-10-28
Consolidate stimulus sequence verification logic

### LINKED ISSUE #528: Headstage64 dialog does not check for valid sequences
The individual device dialogs (`ElectricalStimulatorDialog`) will check if the sequence is valid after the user presses `OK`, but the same check does not occur at the level of the headstage dialog. For consistency, we should either add this check to the headstage dialog, or remove it from the device dialogs, since right now the user could configure the sequence in the headstage, end up with an invalid waveform, save the waveform after closing the headstage dialog, then open the device dialog and without changing anything try to press `OK`, where it will throw a popup window saying the waveform is invalid. This is inconsistent, and should be addressed. 

Question: For consistency, should whatever decision we make also apply to the `Rhs2116` headstage? Currently, the sequence for `Rhs2116` is checked at the headstage level.

bparks13:
This PR moves the stimulus sequence verification logic from the inherited classes to the base class, to increase maintainability. It also adds the verification check when closing the Headstage64 dialog, which confirms that both the Electrical and Optical stimulus sequences are valid before continuing.

Fixes #528

### DIFF
--- a/OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
+++ b/OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
@@ -84,10 +84,35 @@ namespace OpenEphys.Onix1.Design
             }
         }
 
-        internal virtual bool CanCloseForm(out DialogResult result)
+        internal bool CanCloseForm(out DialogResult result, string stimulusName = "Stimulus")
         {
-            result = DialogResult.OK;
-            return true;
+            bool canClose = true;
+
+            if (!IsSequenceValid())
+            {
+                DialogResult resultContinue = MessageBox.Show($"Warning: {stimulusName} sequence is not valid. " +
+                    $"If you continue, the current settings for {stimulusName} will be discarded. " +
+                    "Press OK to discard all changes for this device, or press Cancel to continue editing the sequence.",
+                    $"Invalid {stimulusName} Sequence",
+                    MessageBoxButtons.OKCancel);
+
+                if (resultContinue == DialogResult.OK)
+                {
+                    result = DialogResult.Cancel;
+                }
+                else
+                {
+                    result = DialogResult.OK;
+                    canClose = false;
+                }
+            }
+            else
+            {
+                result = DialogResult.OK;
+            }
+
+            DialogResult = result;
+            return canClose;
         }
 
         internal void OnSelect(object sender, EventArgs e)
@@ -381,7 +406,7 @@ namespace OpenEphys.Onix1.Design
 
         internal virtual bool IsSequenceValid()
         {
-            return true;
+            throw new NotImplementedException();
         }
 
         internal virtual void SetStatusValidity()
--- a/OpenEphys.Onix1.Design/Headstage64Dialog.cs
+++ b/OpenEphys.Onix1.Design/Headstage64Dialog.cs
@@ -1,4 +1,5 @@
-﻿using System.Windows.Forms;
+﻿using System;
+using System.Windows.Forms;
 
 namespace OpenEphys.Onix1.Design
 {
@@ -37,6 +38,20 @@ namespace OpenEphys.Onix1.Design
             OpticalStimulatorSequenceDialog.SetChildFormProperties(this).AddDialogToTab(tabPageOpticalStimulator);
 
             menuStrip1.Visible = false;
+        }
+
+        void OnClickOk(object sender, EventArgs e)
+        {
+            if (ElectricalStimulatorSequenceDialog.CanCloseForm(out DialogResult electricalResult, "Electrical Stimulator")
+                && OpticalStimulatorSequenceDialog.CanCloseForm(out DialogResult opticalResult, "Optical Stimulator"))
+            {
+                if (electricalResult == DialogResult.OK || opticalResult == DialogResult.OK)
+                    DialogResult = DialogResult.OK;
+                else
+                    DialogResult = DialogResult.Cancel;
+
+                Close();
+            }
         }
     }
 }
--- a/OpenEphys.Onix1.Design/Headstage64Dialog.Designer.cs
+++ b/OpenEphys.Onix1.Design/Headstage64Dialog.Designer.cs
@@ -88,7 +88,6 @@
             // buttonOK
             // 
             this.buttonOK.Anchor = ((System.Windows.Forms.AnchorStyles)((System.Windows.Forms.AnchorStyles.Bottom | System.Windows.Forms.AnchorStyles.Right)));
-            this.buttonOK.DialogResult = System.Windows.Forms.DialogResult.OK;
             this.buttonOK.Location = new System.Drawing.Point(881, 2);
             this.buttonOK.Margin = new System.Windows.Forms.Padding(3, 2, 3, 2);
             this.buttonOK.Name = "buttonOK";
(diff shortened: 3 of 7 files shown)


OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
```diff
@@ -84,10 +84,32 @@ void ButtonOk_Click(object sender, EventArgs e)
...
         }
 
-        internal virtual bool CanCloseForm(out DialogResult result)
+        internal bool CanCloseForm(out DialogResult result, string stimulusName = "Stimulus")
         {
-            result = DialogResult.OK;
-            return true;
+            if (!IsSequenceValid())
+            {
+                DialogResult resultContinue = MessageBox.Show("Warning: Stimulus sequence is not valid. " +
```
cjsha: if you want to, you could use `StimulusName` here, like:
`MessageBox.Show($"Warning: {stimulusName} sequence is not...`
but it might be a matter of preference, you ain't gotta take this suggestion.

OpenEphys.Onix1.Design/Headstage64Dialog.cs
```diff
@@ -38,5 +39,18 @@ public Headstage64Dialog(ConfigureHeadstage64 configureNode)
 
             menuStrip1.Visible = false;
         }
+
+        void OnClickOk(object sender, EventArgs e)
+        {
+            if (ElectricalStimulatorSequenceDialog.CanCloseForm(out DialogResult electricalResult, "Electrical Stimulator")
+                && OpticalStimulatorSequenceDialog.CanCloseForm(out DialogResult opticalResult, "Optical Stimulator"))
+            {
+                if (electricalResult == opticalResult)
```
cjsha: i guess I don't understand the logic of allowing the GUIs to close if both stimulus sequences are either both false or both true. If only one of them is invalid, and I click OK, then the GUI doesn't close. I suspect this needs to be revised. maybe something like `if (electricalResult == true && opticalResult== true)`?

OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
bparks13: I like this idea! Thank you

OpenEphys.Onix1.Design/Headstage64Dialog.cs
bparks13: That is a good catch, somehow I never tested that combination of steps. I have now modified it, part of the issue is that I was prioritizing it as an all-or-nothing operation (i.e., both electrical and optical were either discarded or saved), but I realized that I could instead check the return code for each dialog independently, allowing them to correctly close the GUI if needed but still allowing any relevant changes in the other device to be correctly saved.

Please resolve if this behavior is now fixed.

bparks13:
@cjsha I have updated this PR, if you could take a look at it again and see if the behavior now makes sense, that would be great!

OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
```diff
@@ -84,10 +84,35 @@ void ButtonOk_Click(object sender, EventArgs e)
...
+        internal bool CanCloseForm(out DialogResult result, string stimulusName = "Stimulus")
         {
-            result = DialogResult.OK;
-            return true;
+            bool canClose = true;
+
+            if (!IsSequenceValid())
+            {
+                DialogResult resultContinue = MessageBox.Show($"Warning: {stimulusName} sequence is not valid. " +
+                    "If you continue, the current settings will be discarded. " +
```
cjsha: are all settings discarded (not just the stimulation waveform parameters)?

OpenEphys.Onix1.Design/Headstage64Dialog.cs
```diff
@@ -38,5 +39,19 @@ public Headstage64Dialog(ConfigureHeadstage64 configureNode)
...
+
+        void OnClickOk(object sender, EventArgs e)
+        {
+            if (ElectricalStimulatorSequenceDialog.CanCloseForm(out DialogResult electricalResult, "Electrical Stimulator")
+                && OpticalStimulatorSequenceDialog.CanCloseForm(out DialogResult opticalResult, "Optical Stimulator"))
+            {
+                if (electricalResult == DialogResult.OK || opticalResult == DialogResult.OK)
+                    DialogResult = DialogResult.OK;
+                else
+                    DialogResult = DialogResult.Cancel;
```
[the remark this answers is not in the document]
cjsha: I wonder if the logic here can be simplified in general. Like:

```
var estimCloseForm = ElectricalStimulatorSequenceDialog.CanCloseForm(out DialogResult electricalResult, "Electrical Stimulator")
var ostimCloseForm = OpticalStimulatorSequenceDialog.CanCloseForm(out DialogResult opticalResult, "Optical Stimulator"))

if (electricalResult == DialogResult.OK && opticalResult == DialogResult.OK)
    DialogResult = DialogResult.OK;
else 
    DialogResult = DialogResult.Cancel;

if (estimCloseForm && ostimCloseForm)
    Close();
```

or something else like that

OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
bparks13: @cjsha Yes, all settings are discarded.

OpenEphys.Onix1.Design/Headstage64Dialog.cs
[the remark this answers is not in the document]
bparks13: To answer your first question, no, if you choose OK on the first, and Cancel on the second, it will not close the dialog and you have the option to correct the parameters. If the first one is invalid and you click OK, but the second one is valid, the first one will discard parameters but the second one will save parameters. If you click Cancel it will cancel closing the form, which is different from setting the DialogResult to be Cancel.

For your second question, I would prefer not to simplify it in that way, because if either of the results is false, then we shouldn't be setting the DialogResult to anything, since the form is not closing. But there might be a way to simplify it for readability, let me think about it more.

To your third point, I see what you mean, however I am now in the most recent commit using the output of the individual device dialogs to determine the status of the headstage dialog. If they are always set to cancel, then even if there are valid properties that should be saved by one of the devices, if the other is invalid then nothing will be saved, which would be frustrating from a user perspective.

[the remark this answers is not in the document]
cjsha: 1) You're right. I don't know what was going on before. Sorry for the noise.
2) Fair. Although I was confused (and still am a little confused) as to why it works how it does (including the `CanCloseForm()` function), it seems to work well, so don't think about it too much I guess.
3) With the logic from the code block, if there is a valid sequence, that valid sequence would be saved even when discarding an invalid sequence. I just tested (with a couple changes because I used an older commit as the basis as accident), and it seemed to work ok. In any case, in practice, what you already have also seems ok.

This feedback just comes from me trying to parse the code and for some reason I was having difficulty wrapping my head around it (could just be I didn't sleep well last night). Consider the feedback, but feel free to resolve this conversation without taking additional action.

cjsha (approved):
The behavior is good. The two remaining comments are nit picks. One of them involves adding a detail in the messageBox dialog, and the other is just me struggling to understand some logic (though the logic seems to give the desired UI behavior). I'll just approve now so you can just decide yourself if you think it's worth acting on either of these two comments

Issue #515, opened 2025-09-30
Should all properties be represented in the GUIs?


bparks13:
This issue is prompted from [this comment](https://github.com/open-ephys/bonsai-onix1/pull/459#issuecomment-3335472901):

With the addition of the `Enable` properties for the `Headstage64*Stimulator` devices, should we add these (and other) properties to the GUIs? Below are images of the GUIs for these devices, and the Editor in Bonsai showing which properties the user can see:

### Electrical Stimulator

<img width="875" height="518" alt="Image" src="https://github.com/user-attachments/assets/ba9d24fe-4500-4deb-ac8c-071d2eccac83" />

<img width="218" height="244" alt="Image" src="https://github.com/user-attachments/assets/bf28e647-d7b0-4681-95ff-0e5677c91c05" />

Properties missing from the GUI:
- `Enable`
- `StimEnable`
- `PowerEnable`

### Optical Stimulator

<img width="886" height="524" alt="Image" src="https://github.com/user-attachments/assets/36671b55-e0de-475e-8d79-4e48dde080e0" />

<img width="219" height="196" alt="Image" src="https://github.com/user-attachments/assets/64e36f32-6be8-4af6-bb9e-8d4836217d12" />

Missing properties:
- `Enable`
- `StimEnable`

### Rhs2116

<img width="848" height="592" alt="Image" src="https://github.com/user-attachments/assets/6a9cc6fd-68da-44c4-bc26-9bc6f7bf1289" />

<img width="286" height="192" alt="Image" src="https://github.com/user-attachments/assets/5ffa49c2-5764-4841-972c-1209f9c7fcec" />

Missing properties:
- `DspCutoff`
- `AnalogLowCutoff`
- `AnalogLowCutoffRecovery`
- `AnalogHighCutoff`
- `Enable`
- `TriggerSource`
- `Armed`

### Other Devices

Some other devices are also missing properties from the GUI, such as the `NeuropixelsV1e`, which is missing `Enable` and `EnableLed`.

### Questions

- Should we include all additional properties into the GUIs?
- If we are including all properties, how should we move around elements in the existing UI to incorporate these additional properties?
    - One way is that I could convert the existing `Define Stimuli` group box to be a tabbed control instead, with one tab for the stimulus parameters and another tab for other device-specific parameters

PR #531, opened 2025-10-29
Refactor Rhs2116 design files

bparks13:
This PR is to facilitate #515, as the original implementation of the `Rhs2116` design files did not work well with the additions created there. Specifically, when adding a property grid to the `Rhs2116` dialogs, the object that was available was the `Rhs2116StimulusSequencePair`, and not the `ConfigureRhs2116Trigger`. 

This PR updates the `Rhs2116` design files to match the convention of the other design files, where the entire Configure node is passed to the Design library and then individual properties are accessed from there, instead of passing individual properties. By passing the entire configure node, the addition in #515 of a property grid will show the entire `ConfigureRhs2116Trigger` object in the pane, instead of just the stimulus sequence object.

### DIFF
--- a/OpenEphys.Onix1.Design/HeadstageRhs2116Dialog.Designer.cs
+++ b/OpenEphys.Onix1.Design/HeadstageRhs2116Dialog.Designer.cs
@@ -44,14 +44,14 @@
             this.flowLayoutPanel1.SuspendLayout();
             this.SuspendLayout();
             // 
-            // tabControl
+            // tabControlProperties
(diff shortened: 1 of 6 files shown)


aacuevas (approved):
I don't see any problems.
My only suggestion is that, instead of repeating
`var sequence = Trigger.StimulusSequence;` on every method, cache this on the `Trigger` set as a private variable, to avoid so much repetition, since it is something apparently needed everywhere.

PR #534, opened 2025-10-29
Add PropertyGrid to Design dialogs

### LINKED ISSUE #515: Should all properties be represented in the GUIs?

bparks13:
For all custom dialogs, there is now an added tab that contains a PropertyGrid to display all properties for a device. This is always the non-default tab, so that the priority is on the custom GUI, but it is always present should the user want to look at and change any properties that are not exposed via custom controls.

Any changes made in either the custom GUI or the PropertyGrid are operating on the same underlying object, which is the object attached to the PropertyGrid, so that all data is always synchronized.

**Neuropixels V1**
image

**Neuropixels V2**
image

**Rhs2116**
image

**Headstage64**
image

image

Fixes #515


jonnew (changes requested):
As we discussed during live code review, please move the ProbeGroup editor into the the specialized stimulus waveform property editor on the RHS stim gui so that the base class does not have to be informed about its existence.

jonnew:
When some property other than the probeconfiguration is selected, we get a helpful dialog in the area where the probe would be shown. I like this but I want a couple changes.

1. When I resize the form, the text is does not remain centered
    
   image

1. When I select a probe configuration, but I don't have a calibration file loaded, I would like the text to change to "Please provide a calibration file to configure your probe"
   
   image

OpenEphys.Onix1/ConfigureNeuropixelsV2e.cs
```diff
@@ -36,6 +36,15 @@ public ConfigureNeuropixelsV2e(ConfigureNeuropixelsV2e configureNode)
             DeviceAddress = configureNode.DeviceAddress;
         }
 
+        /// <summary>
+        /// Returns a deep copy of the current <see cref="ConfigureNeuropixelsV2e"/> instance.
+        /// </summary>
+        /// <returns><see cref="ConfigureNeuropixelsV2e"/> instance.</returns>
+        public IConfigureNeuropixelsV2 Clone()
```
jonnew: Why are we adding this to the public interface instead of using the copy constructor like all the other configuration operations do? It looks like you are immediately interpreting the result as concrete `ConfigureNeuropixelsV2e` type anyway?

OpenEphys.Onix1/ConfigureHeadstageNeuropixelsV2eBeta.cs
```diff
@@ -37,6 +37,19 @@ public ConfigureHeadstageNeuropixelsV2eBeta()
...
+        /// <summary>
+        /// Initializes a copied instance of the <see cref="ConfigureHeadstageNeuropixelsV2e"/> class.
+        /// </summary>
+        /// <param name="configureNode">Existing <see cref="ConfigureHeadstageNeuropixelsV2eBeta"/> instance.</param>
+        public ConfigureHeadstageNeuropixelsV2eBeta(ConfigureHeadstageNeuropixelsV2eBeta configureNode)
+        {
+            Name = configureNode.Name;
+            Port = configureNode.Port;
+            PortControl = configureNode.PortControl;
+            NeuropixelsV2eBeta = configureNode.NeuropixelsV2eBeta.Clone() as ConfigureNeuropixelsV2eBeta ?? throw new InvalidOperationException($"Unable to copy {nameof(NeuropixelsV2eBeta)} property. Could not cast to the correct type.");
```
jonnew: Something seems wrong here. The most superficial comment is why are you using `as` if you immediately throw an exception if the result is `null`? If you want to throw if the result of `configureNode.NeuropixelsV2eBeta.Clone()` is not castable to ConfigureNeuropixelsV2eBeta, then just use a cast

```
(ConfigureNeuropixelsV2eBeta)configureNode.NeuropixelsV2eBeta.Clone()
```

since this will throw in that case. The second issue is what's up with the `Clone()` method? I have a feeling this was not done at random but I feel like we need to explore if its required or if there is a better way to get the behavior you need.

jonnew (changes requested):
Lets have a live review of this one, I'm not getting the `clone()` stuff.

jonnew (changes requested):
I think we should explore alternatives to this strategy. My big issue with this approach is that its polluting the main library for the sake of a feature we want in the GUIs. The only reason this might be a good idea is that the GUIs are pointing to a desirable feature in the public API that we have not implemented. But, I don't see that happening here. I see these changes as only in service of supporting the specific functionality of the GUI.

So, what's the alternative to get a deep copy for temporarily objects in the GUI? I think one that sounds crazy but actually is quite elegant is to use the Json serilize that you already have access to as a dependency. Something like:

``` csharp
using Newtonsoft.Json;

namespace OpenEphys.Onix1.Design
{
    internal static class DeepCopyHelper
    {
        public static T SerializeDeserialize<T>(T source)
        {
            var json = JsonConvert.SerializeObject(source);
            return JsonConvert.DeserializeObject<T>(json);
        }
    }
}
```
and 

``` csharp
 var configuration = DeepCopyHelper.SerializeDeserialize(configureHeadstage);
```

(using `NeuropixelsV1eHeadstageEditor` as an example) seems like it might fulfil your requirements without any modifications to the base library at all.

PR #548, opened 2025-11-12
Set Rhd2164 DspCutoff to Off

### LINKED ISSUE #543: Default DSP should be set to off in ConfigureRhd2164
For some reason, the default value for the DSP setting in `ConfigureRhd2164` is set to 0.146 Hz. 

https://github.com/open-ephys/bonsai-onix1/blob/7fb9332eb67c111a894de5ecaddbe63def120562/OpenEphys.Onix1/ConfigureRhd2164.cs#L55

It should be set to off to match the analog settings, which default to the widest pass band possible.

bparks13:
To give the widest pass band possible, the DSP cutoff value is now set to Off.

Fixes #543

### DIFF
--- a/OpenEphys.Onix1/ConfigureRhd2164.cs
+++ b/OpenEphys.Onix1/ConfigureRhd2164.cs
@@ -51,7 +51,7 @@ namespace OpenEphys.Onix1
         /// Gets or sets the cutoff frequency for the digital (post-ADC) high-pass filter used for amplifier offset removal.
         /// </summary>
         /// <remarks>
-        /// The amplifiers on the Intan chip, past the analog filter, introduce a DC offset that varies with each channel.
+        /// The amplifiers on the RHD2164 chip, past the analog filter, introduce a DC offset that varies with each channel.
         /// The <see cref="Rhd2164DspCutoff"/> exists to remove this DC offset and ensure that all signals are centered at zero.
         /// With it disabled, all the signals will appear centered at different values.
         /// </remarks>


aacuevas:
I must say, I'm not 100% convinced defaulting the DSP to disabled is the best option
Just as a reminder of what de DSP does: The amplifiers on the Intan chip, past the analog filter, introduce a DC offset that varies with channel. The DSP exists to remove this DC offset and ensuring that all signals are 0-centered. With it disabled, all the signals will appear centered about different values.

Most analysis do their own filtering afterwards, so it is true that in most cases this just adds an unnecessary additional FIR filtering (with all its potential effects). But it will be very noticeable when plotting the signals. 

So if we set this to off by default, we need to thoroughly document this.

jonnew:
I understand but having it off is in line with the the least filtering of data by default principle that we tend to follow. e.g. the analog settings are both maxed out.

What does to OE GUI acq. board plugin do by default?

In any case, the text in your comment can simply go in the XML description of this setting!

OpenEphys.Onix1/ConfigureRhd2164.cs
```diff
@@ -52,7 +52,7 @@ public ConfigureRhd2164(ConfigureRhd2164 configureRhd2164)
         /// </summary>
```
bparks13: Add details to the XML comment specifying a remark:

The amplifiers on the Intan chip, past the analog filter, introduce a DC offset that varies with channel. The DSP exists to remove this DC offset and ensuring that all signals are 0-centered. With it disabled, all the signals will appear centered about different values.

PR #552, opened 2025-11-17
Add support for NPX 2.0 single-shank probes

### LINKED ISSUE #446: Add support for single-shank NP 2.0 probes
Single shank NP 2.0 are available for purchase: https://www.neuropixels.org/probes-np2-0. We've had user requests about using them with ONIX.

@jsiegle would you be able to let us know how they work with our NP2.0e headstages?
Then the dev team can evaluate including them in the configuration GUI.

bparks13:
This PR supersedes #493.

Adding support for single-shank Neuropixels 2.0 probes, with the additional functionality of allowing the user to change the probe type in the GUI.

> Note that due to the restructuring of the abstract/concrete classes, in the XML file the name of the property is `ProbeConfiguration` for the quad-shank probe (this is simply the name of the property, and is what all older versions of the files will be named), but for the single-shank probe it is renamed to `SingleShankProbeConfiguration`. We could keep the same name (`ProbeConfiguration`) and add the type name in the XML tag (i.e., `xsi:type="onix1:SingleShank"`), but this will break backward compatibility, since the deserializer cannot know which concrete type to use for the probe configuration property when loading old workflows. If we want to break compatibility for this, we can inform users that they can add the line `xsi:type="onix1:QuadShank"` to the ProbeConfiguration properties to let it load, but it seemed easier to maintain backward compatibility.

The text above is now outdated, with the most recent commits there is now a non-browsable, non-externalizable property that is used to handle serialization, leveraging the attributes to automatically add the type; this is called `ConfigurationA`/`B`. Additionally, the original `ProbeConfigurationA`/`B` properties are not serialized, but they are deserialized into a quad-shank probe configuration, allowing for backward compatibility.

@anjaldoshi We do not have any single-shank probes to test with, could you build this branch and run some recordings with a single-shank probe to make sure it records data correctly?

Fixes #446

### DIFF
--- a/OpenEphys.Onix1/NeuropixelsV2eProbeGroup.cs
+++ b/OpenEphys.Onix1/NeuropixelsV2eProbeGroup.cs
@@ -87,7 +87,7 @@ namespace OpenEphys.Onix1
         private static float ContactPositionX(int index)
         {
             var shank = index / NeuropixelsV2.ElectrodePerShank;
-            var offset = ShankOffsetX + (ShankWidthX + ShankPitchX) * shank + 11;
+            var offset = ShankOffsetX + ShankPitchX * shank + 11;
 
             return (index % 2) switch
             {
--- a/OpenEphys.Onix1/ConfigurePersistentHeartbeat.cs
+++ b/OpenEphys.Onix1/ConfigurePersistentHeartbeat.cs
@@ -28,7 +28,6 @@ namespace OpenEphys.Onix1
         /// <summary>
         /// Gets or sets the rate at which beats are produced in Hz.
         /// </summary>
-        [Range(100, 10e6)]
         [Category(AcquisitionCategory)]
         [Description("Rate at which beats are produced (Hz).")]
         public uint BeatsPerSecond
--- a/OpenEphys.Onix1/ConfigureRhd2164.cs
+++ b/OpenEphys.Onix1/ConfigureRhd2164.cs
@@ -50,9 +50,14 @@ namespace OpenEphys.Onix1
         /// <summary>
         /// Gets or sets the cutoff frequency for the digital (post-ADC) high-pass filter used for amplifier offset removal.
         /// </summary>
+        /// <remarks>
+        /// The amplifiers on the RHD2164 chip, past the analog filter, introduce a DC offset that varies with each channel.
+        /// The <see cref="Rhd2164DspCutoff"/> exists to remove this DC offset and ensure that all signals are centered at zero.
+        /// With it disabled, all the signals will appear centered at different values.
+        /// </remarks>
         [Category(ConfigurationCategory)]
         [Description("Specifies the cutoff frequency for the digital (post-ADC) high-pass filter used for amplifier offset removal.")]
-        public Rhd2164DspCutoff DspCutoff { get; set; } = Rhd2164DspCutoff.Dsp146mHz;
+        public Rhd2164DspCutoff DspCutoff { get; set; } = Rhd2164DspCutoff.Off;
 
         /// <summary>
         /// Gets or sets the low cutoff frequency of the analog (pre-ADC) bandpass filter.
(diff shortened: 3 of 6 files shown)


OpenEphys.Onix1.Design/ChannelConfigurationDialog.cs
```diff
@@ -311,10 +311,8 @@ internal bool ValidateProbeGroup(ProbeGroup newConfiguration)
             }
             else
             {
-                MessageBox.Show($"Error: Number of contacts does not match; expected {ProbeGroup.NumberOfContacts} contacts" +
-                    $", but found {newConfiguration.NumberOfContacts} contacts", "Contact Number Mismatch");
-
-                return false;
+                throw new InvalidOperationException($"Number of contacts does not match; expected {ProbeGroup.NumberOfContacts} contacts" +
+                    $", but found {newConfiguration.NumberOfContacts} contacts. Ensure the file contains the correct probe group type.");
```
jonnew: "probe group" is coming from probe interface file, right? It sounds weird from a normal language perspective.

OpenEphys.Onix1/NeuropixelsV2ProbeConfiguration.cs
```diff
@@ -217,6 +238,32 @@ public string ProbeInterfaceFileNameSerialize
...
+        const int ReferencePixelCount = 4;
+        const int DummyRegisterCount = 4;
+
+        /// <summary>
+        /// Number of registers per shank.
+        /// </summary>
+        protected const int RegistersPerShank = NeuropixelsV2.ElectrodePerShank + ReferencePixelCount + DummyRegisterCount;
+
+        /// <summary>
+        /// Index of the shift register bit for external electrode 0.
```
jonnew: xml comments are not needed for protected members. Further these comments do not provide information over the variable name.

jonnew (changes requested):
I think there are some pretty strong anti patterns in here. I would take a look at those and more broadly review your work to make sure that your following OOP best practices.

From a UX perspective, its good.

OpenEphys.Onix1/NeuropixelsV2ProbeConfiguration.cs
bparks13: Removing the comment for protected members leads to a compiler warning, as it is considered a publicly visible member. I added simple comments to remove the warning. However, I can make the constants `internal` instead to restrict their usage but also remove the comments.

OpenEphys.Onix1/NeuropixelsV2eProbeGroup.cs
```diff
@@ -50,6 +65,74 @@ internal void UpdateDeviceChannelIndices(NeuropixelsV2Electrode[] channelMap)
...
+
+            for (int i = 0; i < numberOfContacts; i++)
+            {
+                var shank = i / NeuropixelsV2.ElectrodePerShank;
+                contactIds[i] = shank switch
+                {
+                    0 => "0",
+                    1 => "1",
+                    2 => "2",
+                    3 => "3",
```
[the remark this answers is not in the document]
bparks13: It does apply to both quad- and single-shank, but I agree that it is confusing to leave it in the base class since the single-shank simply returns `"0"` for all contacts. I removed it from the base class for consistency and clarity.

jonnew (changes requested):
Almost there.

OpenEphys.Onix1/NeuropixelsV2QuadShankProbeConfiguration.cs
```diff
@@ -47,7 +47,7 @@ public enum NeuropixelsV2QuadShankReference : uint
     [XmlType(TypeName = XmlTypeName, Namespace = Constants.XmlNamespace)]
     public class NeuropixelsV2QuadShankProbeConfiguration : NeuropixelsV2ProbeConfiguration
     {
-        const string XmlTypeName = "QuadShank";
+        internal const string XmlTypeName = nameof(NeuropixelsV2QuadShankProbeConfiguration);
```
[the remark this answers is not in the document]
bparks13: It is internal because I previously was using the type name as the name of the property in the Xml serializer; however, we decided that it was better to have a consistent name for the property, and so this was no longer needed outside of this class. I failed to remove it though, so thank you for catching this.

PR #529, opened 2025-10-24
Add support for headstage-64 firmware version 0.4

jonnew:
- This PR supersedes #521 
- Fixes #502 
- Fixes #516 (supersedes #520)
- Fixes #537
- Fixes #542

PR #564, opened 2026-01-12
Disable showing spurious channel numbers for stimulus sequence dialogs

### LINKED ISSUE #540: Rhs2116 stimulus sequence shows channel numbers above the 32-channel range
First referenced [here](https://github.com/open-ephys/bonsai-onix1-docs/pull/276#pullrequestreview-3409169064), the Rhs2116 stimulus sequence dialog shows channel numbers above the listed 32 channels, and includes dashed lines for them. This is not the case below the channel range, as we have ZedGraph cut off the y-axis labels below the last channel. The quirk with ZedGraph is that we cannot easily do this operation on both sides.

This was not a consideration until recently, as we used to restrict the viewport to only show these 32 channels, but as of #497 we no longer are restricting this view, which leads to being able to see the labels above the valid range.

Is this worth working on? Should we leave it as-is?

bparks13:
This PR fixes #540, by removing any labels outside of the valid range of channel numbers. Additionally, when this was first implemented in #529, I also implemented a fix for zooming in and out of the stimulus sequence dialogs. That fix is also implemented here in a separate commit.

image

Showing how even with vertical zoom disabled, everything is still centered on the mouse pointer
zooming


jonnew (approved):
Looks good.

PR #568, opened 2026-01-26
Combine HS64 electrical stimulator's Enable / PowerEnable properties

### LINKED ISSUE #511: Combine Headstage64ElectricalStimulator's Enable and PowerEnable properties
There are de-facto two parallel enables for Headstage64ElectricalStimulator. 

1. `Enable`: a logical enable that controls the state machine that generates stimulus sequences
2. `PowerEnable`: A 1 bit signal sent to the high compliance voltage power supplies. 

These should be combined into a single `StimulatorEnable` property for this library. Currently, if `Enable` is true but `PowerEnable` is false, a stim train will be generated and applied to the analog current source, but this will not be properly powered resulting in undefined behavior at the stimulus pin. Thus if the user wants to enable the stimulator, the power should be turned on. It should be noted in the documentation that it takes some time for the power supply to stabilize, so stimuli cannot be properly delivered directly after enabling. Ill take a measurement on how long this takes and append to this issue.

bparks13:
This issue was originally resolved in PR #529, but it also involved a firmware update which convoluted the cherry-picking process. To ensure that we have the behavior we want, but also staying consistent with the most recent version of the library, this PR is merging into the PR that reverts Headstage64 stimulus reports so that the naming scheme is consistent.

Here, we are combining the current `Enable` / `PowerEnable` properties so that they are presented to the user as a single parameter `StimulatorEnable`.  

Fixes #511

### DIFF
--- a/OpenEphys.Onix1/ConfigureHeadstage64ElectricalStimulator.cs
+++ b/OpenEphys.Onix1/ConfigureHeadstage64ElectricalStimulator.cs
@@ -19,7 +19,7 @@ namespace OpenEphys.Onix1
     [Editor("OpenEphys.Onix1.Design.Headstage64ElectricalStimulatorComponentEditor, OpenEphys.Onix1.Design", typeof(ComponentEditor))]
     public class ConfigureHeadstage64ElectricalStimulator : SingleDeviceFactory
     {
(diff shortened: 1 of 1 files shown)


jonnew:
This is already fixed in #569, I think this commit is redundant.

PR #571, opened 2026-01-29
Modify headstage-64 port voltage scan algorithm

jonnew:
- I've made several changes to the algorithm to get functionality with tethers of different lengths and different POR behavior across headstages
- There arei two major changes:
  1. The scan starts with the application of 10V for 10-20 msec. This is long enought to turn on the LDOs across tethers and short enought not to result in overvoltage damge for short tethers
  2. Apply an offset voltage that is a linear function of the minimum lock voltage instead of a constant. m and b were found emperically using 0.7 and 2.5 m tether in order to get 5.2 at the headstage.

### DIFF
--- a/OpenEphys.Onix1/ConfigureHeadstage64.cs
+++ b/OpenEphys.Onix1/ConfigureHeadstage64.cs
@@ -175,8 +175,6 @@ namespace OpenEphys.Onix1
 
                 // NB: Wait for 1 second to discharge the headstage in the case that they have e.g. just
                 // restarted the workflow automatically with nearly no delay from the last run.
-                //device.WriteRegister(PortController.PORTVOLTAGE, (uint)(MinVoltage * 10));
-                //Thread.Sleep(500);
                 device.WriteRegister(PortController.PORTVOLTAGE, 0);
                 Thread.Sleep(1000);
 


jonnew:
@bparks13 Aside from a code review can you help get @ChucklesOnGitHub set up with a nuget package to check these changes with OEPS hardware. @ChucklesOnGitHub after you let me know the result of that we can bring this algorithm over to the update software (or not if its unreliable).

OpenEphys.Onix1/ConfigureHeadstage64.cs
```diff
@@ -168,16 +168,23 @@ protected override bool ConfigurePortVoltage(DeviceContext device, out double vo
...
-                const double MaxVoltage = 6.0;
-                const double VoltageOffset = 4.0;
+                const double MaxVoltage = 6.5;
                 const double VoltageIncrement = 0.2;
 
                 // NB: Wait for 1 second to discharge the headstage in the case that they have e.g. just
-                // restarted the workflow automatically with nearly no delay from the last run. 
+                // restarted the workflow automatically with nearly no delay from the last run.
+                //device.WriteRegister(PortController.PORTVOLTAGE, (uint)(MinVoltage * 10));
+                //Thread.Sleep(500);
```
bparks13: Are these supposed to be commented out?

PR #583, opened 2026-02-26
Add PropertyGrid to design dialogs

### LINKED ISSUE #515: Should all properties be represented in the GUIs?

bparks13:
This PR supersedes #534, as there was a lot of back-and-forth in the commit history.

The functionality in this branch is the same as in the original branch, from a user perspective. On the backend, all copy constructors that were only added for use by the Design library have now been removed, and instead a copy helper method has been added to the DesignHelper class. There is also a deep copy helper method that is used to ensure that any changes made in the dialogs can be correctly rolled back if the user cancels a dialog.

Fixes #515


cjsha (approved):
EDIT: I approved changes accidentally, I meant to add submit approval for a different PR.

cjsha (changes requested):
how do I switch between editing probeinterface file for probe A and probeinterface file for probe B here?

image

Afaik, there's no way to tell which one I'm editing. This issue seems to pertain to all np headstages that accept more than one probe except for v1f wherein each probe is its own device, I believe

---

This doesn't seem implemented for v1e or v1f.

image

---

visual nitpick: why not keep it on the the property editor on the left side for all tabs? for example, it's on the left under the NeuropixelsV2eBeta tab and it's on the right under the Bno055 tab. If this is something we want to change, it's probably beyond the scope of this PR and should belong in a separate issue.

image
image

cjsha (approved):
This looks good to me. I made #601 for us to discuss possible future changes that might not be worth delaying 0.7 release

PR #588, opened 2026-03-04
Use ProbeInterface files for configuration directly

### LINKED ISSUE #580: ProbeInterface updates
Implementing the proposal from Bruno / Goncalo to only use the ProbeInterface filename instead of the ProbeGroup property.

TODO: add more details here

bparks13:
This PR aims to remove all `ProbeGroup` properties from configuration classes, and instead solely rely on the configuration filename to load the `ProbeGroup` whenever a GUI is opened, or when the workflow is run. This way there is no confusion about which configuration is loaded, simplifying the class API surface as well by removing file I/O operations that occurred during serialization/deserialization.

The dialogs have been tested, and should be ready for users to interact with them. We have moved the File menu so that it is clearer which tab the menu strip is operating on, and also added Save / Save As... functionality. The current state of the electrode configuration (i.e., when it is dirty) is denoted by an asterisk on either the title of the window or the tab, whichever is more appropriate. 

On hardware, this has been tested with a `HeadstageNeuropixelsV1e` and a `HeadstageNeuropixelsV2eBeta`; the same code should work for the V1f/V2e variants, but should be tested before releasing.

Fixes #580

### DIFF
--- a/OpenEphys.Onix1.Design/ChannelConfigurationDialog.Designer.cs
+++ b/OpenEphys.Onix1.Design/ChannelConfigurationDialog.Designer.cs
@@ -35,7 +35,11 @@
             this.menuStrip = new System.Windows.Forms.MenuStrip();
             this.fileMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.dropDownOpenFile = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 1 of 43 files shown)


cjsha (approved):
seems good to me

PR #591, opened 2026-03-05
Move raw serdes config to headstage level

### LINKED ISSUE #165: Rethink enable logic for NPX logic and possibly other passthrough devices
Today during testing I realized how annoying this error is for quick debugging since I didn't have any calibration file handy. Would it be possible to either specify a default value and ignore the file entirely?

https://github.com/neurogears/onix-refactor/blob/68552dfd62b56828b93ed432710dbd345bcb64e7/OpenEphys.Onix/OpenEphys.Onix/ConfigureNeuropixelsV2eBeta.cs#L169-L172

Furthermore, and more to the topic of this issue, it hit me that these passthrough headstages work as a mono unit, in that we cannot configure for example the NPX v2 BNO055 device without also configuring the NPX v2 node. Because of all the configuration and validation / error checking that goes into setting up these devices through the passthrough, we should consider breaking out of the configuration logic early if `Enable` is set to false, so that none of these checks (probes detected, calibration files, etc) would be triggered if the probe is disabled.

This would be invaluable to allow testing for example the BNO055 without needing an actual probe connected to the headstage.

### LINKED ISSUE #468: Rethink Enable logic on ConfigureNeuropixelsVXe operators
`ConfigureNeuropixelsVXe` operators expose an `Enable` property.  This is tied to the `DS90UB9x.ENABLE` register state. 

There is also a per-probe, implicit enable property as well:

https://github.com/open-ephys/bonsai-onix1/blob/b9bdc3c0bb340843ff4288b6145308176307d81e/OpenEphys.Onix1/ConfigureNeuropixelsV2e.cs#L180

If the probe is not detected, then its configuration is skipped and it is de-facto disabled. This is a bit of a logical mess

1. If `Enable` is set to `false`, then why are detected probes being configured? They will ultimately produce no data.
2. Even worse, if the user sets `Enable` to false, presumably to test something about acquisition without probes plugged in, then both `ProbeSerialNumber`s will be null and exception will be thrown:

    https://github.com/open-ephys/bonsai-onix1/blob/b9bdc3c0bb340843ff4288b6145308176307d81e/OpenEphys.Onix1/ConfigureNeuropixelsV2e.cs#L165

   why are we requiring probes to be present when their data will be ignored?

3. Finally, as a result of playing around with this issue, I found that if probe configuration is skipped then 9 times of 10, this line will through a Failure to write register exception:
   https://github.com/open-ephys/bonsai-onix1/blob/b9bdc3c0bb340843ff4288b6145308176307d81e/OpenEphys.Onix1/ConfigurePolledBno055.cs#L120

   Putting a `Thread.Sleep(100);` before this line "solves" this issue. It seems that we were getting away with this because configuring the Neuropixels, which is currently required, takes a while.

So this is a bit of a mess. At a minimum, if the `Enable` register is false, the minimal amount of setup should be performed while leaving the probes in powered down state even if they are present.  Further, in this case, it should be acceptable for no probes to be detected since the user has indicated they don't want their data anyway.

### LINKED ISSUE #590: Configuration of raw deserializer for passthrough devices should be at headstage level
For headstages that use the passthrough deserialzer to transmit data, deserializer configuration is handled piecemeal by individual "virtual" devices with a complete lack of required coordination. For example, in the `NeuropixelsV2e `, the `ConfigureNeuropixelsV2e` and `Bno055` properties each handle a portion of deserializer configuration without coordination, potentially overwriting each others actions. e.g., consider if the I2C aliases happened to clash. 

The solution to this is for these properties and their associated classes to be be as close to "virtual" devices as possible with all real hardware setup taken care of at the level of the headstage. This is an extremely similar pattern to the abstract `ConfigurePortController`, which is implemented at the headstage level in order to achieve a headstage specific port voltage controller.

jonnew:
- This draft commit moves all raw SERDES configuration, which requires coordination for various supprted devices within a MultiDeviceFactory to the MultiDeviceFactory level
- It introduces an abstract ConfigureDS90UB9x class that can be implmented in MultiDeviceFactories to configurure the Serdes for a particular headstage
- ConfigureHeadstageNeuropixelsV2e uses this class and can serve as a template for other MultiDeviceFactories that would benefit: ConfigureHeadstageNeuropixelsV2eBeta, ConfigureHeadstageNeuropixelsV1e, and ConfigureMiniscopeV4
- Because this change almost completely abstracts hardware configuration from the devices within the Multidevice factory, simplifications can be made to those devices. For instance, ConfigureHeadstageNeuropixelsV2e now contains devices for each probe instead of a single device that combines the streams from each. This is a much more idiomatic approach, but has downstream affects on Data nodes and the visualization library. The later was not addressed in this commit.
- Fixes #590 
- Fixes #468
- Fixes #165 (to some extent. The calibration file is required because operation without it is dangerous enough to data integrity its not something we want to allow)


jonnew:
@bparks13 this is now ready for you to take over with respect to design library changes and main merge and or rebasing.

bparks13:
@jonnew The dialogs are now fully functional again, and this PR is ready for integration testing. 

I went through and checked all of the major functionality for the dialogs that was introduced in v0.7.0, and everything is working as expected. 

The biggest change in terms of dialogs is how the Neuropixels 2.0 probes are presented to the user; now that only one probe configuration exists per `NeuropixelsV2PsbDecoder`, there is now one tab per probe configuration. This matches the pattern that exists in all other headstage dialogs, and also reflects the underlying class architecture.

jonnew:
Great, Ill pull these down and have a look.

jonnew:
Ok, this looks good. I added the GUI for the new headstage. @bparks13 Please perform final review before I merge.

OpenEphys.Onix1/HeadstageEeprom.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Text;
+
+namespace OpenEphys.Onix1
+{
+    class HeadstageEeprom
+    {
+        public const uint I2CAddress = 0x51;
+
+        // header field offsets (absolute, in bytes)
+        const uint HeaderIdentOffset = 0; // 4 bytes: "OEHS"
```
bparks13: Comment says "OEHS", but lines 46-54 expect it be "HSOE".

bparks13 (approved):
I tested the new headstage dialog, pushed some fixes to the behavior so it is consistent with all other dialogs. Fixed a couple XML comments as well.

Everything else looks good, the only question I have is regarding the comment below. Let me know if I need to get some hardware connected and test these changes, or if the integration testing has already been done.

OpenEphys.Onix1/HeadstageEeprom.cs
jonnew: fixed.

PR #595, opened 2026-03-05
Update Stimulus Sequence dialogs

### LINKED ISSUE #593: Update stimulus sequence dialogs
Due to the changes made in #588, to implement #580, the `ChannelConfigurationDialog` was updated. One way it was updated is by changing how the menu strip is handled; this includes the items in the File menu, how the items are nested, and where the menu strip appears within a given dialog. The default behavior of that dialog now leaves the menu strip at the level of the dialog that created it, instead of propagating the menu items up to the top-level parent; this is important for the `GenericStimulusSequenceDialog`, because it can contain a nested `ChannelConfigurationDialog`, such as for a `Rhs2116StimulusSequenceDialog`. 

Therefore, we need to update how the menu strip is handled, for the generic dialog as well as for the concrete dialogs such as `Headstage64` and `HeadstageRhs2116`. One specific update for `HeadstageRhs2116` is to remove the menu strip from the `ChannelConfigurationDialog`, and add a menu item to the stimulus sequence menu strip that says "Import Probe Configuration". The reason is because we are not changing anything about the ProbeGroup that can or should be saved back to the file, therefore we only need an import option.

Additionally, the current nesting for the File menu should be flattened, and the options renamed to "Open Stimulus Waveform" and "Save Stimulus Waveform".

bparks13:
Building off of the changes introduced in #588, this PR removes the `ProbeGroup` and `ProbeInterfaceFileName` from the `Rhs2116Trigger` class; the probe group is only used for visualization, and should not be part of the class. Related changes are then propagated to the `Rhs2116StimulusSequenceDialog` class, where the default `ProbeGroup` is loaded for visualization.

Updated the `GenericStimulusSequenceDialog` to match the Menu strip changes in `ChannelConfigurationDialog`, such as only having the menu strip at the level of the tab that the menu items are operating on. This PR also removes the menu strip from the channel configuration dialog that the Rhs2116 dialog shows, and moves the Import menu item to be part of the menu strip for the Trigger tab for consistency.

Fixes #593


cjsha (approved):
The stimulus UI behavior seems ok to me.

---

I know this is the beyond the scope of this PR, but do we want waveform GUIs to be views into a file like we do for probe configuration?

PR #598, opened 2026-03-10
Add minimum version to NeuropixelsV1f

bparks13:
During testing, we discovered a potential issue after PR #512 was merged; the `NeuropixelsV1f` device did not have a minimum version, leading to an error when acquisition is started:

image

This PR adds in the minimum version to the static class, but then we get a secondary error:

image

It appears that the BNO on this device is outdated. If this is only an issue of the BNO on this particular device, then we can review and merge this PR; however, if all V1f headstages all contain the older version of the BNO, v0.7.0 would mean that all V1f devices would no longer be able to record data.

### DIFF
--- a/OpenEphys.Onix1.Design/NeuropixelsV1fHeadstageEditor.cs
+++ b/OpenEphys.Onix1.Design/NeuropixelsV1fHeadstageEditor.cs
@@ -6,7 +6,7 @@ using System;
 namespace OpenEphys.Onix1.Design
 {
     /// <summary>
-    /// Class that opens a new dialog for a <see cref="ConfigureHeadstageNeuropixelsV1e"/>.
+    /// Class that opens a new dialog for a <see cref="ConfigureHeadstageNeuropixelsV1f"/>.
     /// </summary>
     public class NeuropixelsV1fHeadstageEditor : WorkflowComponentEditor
     {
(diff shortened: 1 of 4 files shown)


bparks13:
@jonnew @aacuevas @cjsha Cris and I have been workshopping what to do about V1f, and we have settled on the following changes:

- Mark all `NeuropixelsV1f` classes as `Obsolete`, indicating that these nodes are no longer supported in v0.7.0
  - This effectively hides the nodes from being added to a workflow. It does not hinder the user from opening an old workflow which contains a V1f node in it, but that node will have cross-hatches on it indicating its deprecated status
- Downgrade the minimum version of the `Bno055` and `TS4231` devices to be version 1 
  - (Optional) If we do not do this, then the `NeuropixelsV1f` headstage will be non-functional in v0.7, as the `Bno055` and `TS4231` devices are not running a supported device version. 

These are the minimal changes we converged on that still allowed the existence of functional V1f nodes in v0.7.0.

We also attempted to use the device configure nodes as a workaround, to only configure the `NeuropixelsV1f` itself, but could not get it to work due to the port voltage not being turned on since we no longer have a headstage node configuring the port.

cjsha:
Alternatively, what about trying to update the v1f? I know the hs64 had problems but maybe other headstages can be updated. I propose delaying obsoleting until we know it can't be updated? 

In this case, the PR would only decrement the version requirements, no obsoleting.

bparks13:
From issue sorting meeting:

- @cjsha suggested not marking the V1f as obsolete, since we can always do that later.
- @jonnew suggested adding an issue where V1 of Bno/Ts4231 are being kept in the library due to V1f, and will be updated in a future release. The V1f will not be marked as obsolete. 
  - Target version 1.0

cjsha (approved):

PR #613, opened 2026-03-25
Allow Neuropixels probes to run without a ProbeInterface file

bparks13:
- Loads the default configuration, which is typically the first 384 channels enabled
- For NeuropixelsV2e, the default probe type is the quad-shank 2.0 probe


jonnew (approved):
This looks good.

PR #617, opened 2026-04-01
FrameWriter

### LINKED ISSUE #247: Proposal: DataFrameWriter
Sometimes we use `enum`s as members of data frames:

https://github.com/open-ephys/onix-bonsai-onix1/blob/e5e5a97a956b60e4bd04fa9915c5774278579f90/OpenEphys.Onix1/BreakoutDigitalInputDataFrame.cs#L26

These can be nice because they can be used with `HasFlag` to determine if they contain some value (e.g. a pin is high or button was pressed in this case). However, in the context of Bonsai's existing infrastructure, they are basically dead-on-arrival when it comes to saving data. We have two options that I can see, currently:

1. Use `CsvWriter`: This option is not good. We can look some `;`-delimited example data produced by the breakout board to see why:

```
...
4410751;Pin0, Pin1, Pin2, Pin3, Pin4, Pin5;PortBOn, PortAOn
4410813;Pin0, Pin1, Pin2, Pin3, Pin4, Pin5, Pin6, Pin7;PortBOn, PortAOn
4416751;Pin0, Pin1, Pin2, Pin3, Pin4, Pin5;PortBOn, PortAOn
4416813;Pin0, Pin1, Pin2, Pin3, Pin4, Pin5, Pin6, Pin7;PortBOn, PortAOn
...
```
- The first element is a `ulong`
- The second element is all the bit flags within the `BreakoutDigitalPortState` that are high. In reality, this variable sized, human readable array with the word "Pin" repeated over and over again is encoding a byte. Not smart. Not only extremely inefficient but a complete nightmare to even parse in a reasonable way.
- The last element is the same story for the state of buttons and switches on the breakout.

2. We could split out each element of the data frame and use a `MatrixWriter`. The caveat is that `MatrixWriter` requires an array of privative types. So for each element of the data fram we must convert the enum to its underlying type using an `ExpressionTransform`: `byte(it)`, then `BufferCount` the byte into a single  or multielement Enumerable, then convert to an array, and then use matrix writer. Not good. There is some hope in this issue: https://github.com/bonsai-rx/bonsai/issues/1578. In that case we could just use `BinaryWriter` (or whatever it ends up being called) directly on the enums, but we don't know the timeline. 

The real issue here is probably bigger than just this: I think we need some type of generic `DataFrameWriter` sink that can go right after any of the `XXXData` nodes to serialize their streams to disk in a reasonable way, rather than having to split out elements before saving. I know that @glopesdev had something for serializing `oni.frame` data to individual files based on the device address, but this is too low level IMO. The `XXXData` nodes often perform a good deal of reorganization and processing that would need to be redone offline.

Here is a non-functional idea. This code compiles, but for reasons I don't yet understand, Bonsai will not load the operator. We may just need to make a concrete `DataFrameWriter` for every correspond `DataFrame` type instead. Maybe we could banish this atrocity to its own `OpenEphys.Onix1.Utils` package.

``` csharp
public abstract class DataFrame
{
    internal DataFrame(ulong clock)
    {
        Clock = clock;
    }

    internal DataFrame(ulong clock, ulong hubClock)
        : this(clock)
    {
        HubClock = hubClock;
    }

    public ulong Clock { get; }

    public ulong HubClock { get; internal set; }

    internal virtual void Write(BinaryWriter bw)
    {
        bw.Write(Clock);
        bw.Write(HubClock);
    }
}

  [Description("Records frame.")]
  public class DataFrameWriter<Tsource> : Sink<Tsource> where Tsource : DataFrame
  {
        const string fn = "./test.raw";

        public override IObservable<Tsource> Process(IObservable<Tsource> source) 
        {
            return Observable.Using(
                () => new BinaryWriter(new FileStream(fn, FileMode.Create)),
                bw => source.Do(f =>
                {
                    f.Write(bw);
                }).Finally(bw.Close));
        }
  }
```

bparks13:
This PR is to add the Onix1.FrameWriter as a potential new method for saving data frames. Any of the data frames (or buffered data frames) from Onix1 can be an input to the node. Reflection and expression trees are utilized to build the conversion from a frame to an Arrow array, which is then written to the file.

This uses `Apache.Arrow` as the file writer. 

Fixes #247

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <PackageIcon>icon.png</PackageIcon>
-    <VersionPrefix>0.7.0</VersionPrefix>
+    <VersionPrefix>0.8.0</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <LangVersion>10.0</LangVersion>
     <Features>strict</Features>
--- a/OpenEphys.Onix1/Bno055DataFrame.cs
+++ b/OpenEphys.Onix1/Bno055DataFrame.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Numerics;
 using System.Runtime.InteropServices;
+using OpenEphys.Onix1.FrameWriter;
 
 namespace OpenEphys.Onix1
 {
@@ -76,6 +77,7 @@ namespace OpenEphys.Onix1
         /// <summary>
         /// Gets the 3D orientation represented as a Quaternion.
         /// </summary>
+        [FrameWriterIgnoreMembers(MemberType.Properties)]
         public Quaternion Quaternion { get; }
 
         /// <summary>
--- a/OpenEphys.Onix1/FrameWriter/ArrowBatchWriter.cs
+++ b/OpenEphys.Onix1/FrameWriter/ArrowBatchWriter.cs
@@ -1,0 +1,74 @@
+﻿using System;
+using System.Collections.Generic;
+using Apache.Arrow;
+
+namespace OpenEphys.Onix1.FrameWriter
+{
+    /// <summary>
+    /// Provides buffered writing of items to an Arrow file, batching items before writing.
+    /// </summary>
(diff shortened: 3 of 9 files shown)


bparks13:
Superseded by #625

PR #620, opened 2026-04-13
Decrease minimum version check for Rhd2164

### LINKED ISSUE #618: Rhd2164 device in the field fails minimum version check
0.7.0 implemented a version check. 

During internal discussions, we found that the v2 of the Rhd2164 device actually failed to adhere to our [spec](https://open-ephys.github.io/ONI/v1.0/hw-spec/devices.html#device-descriptor):

> Device_Version: A version number to distinguish between implementations of a singular device type. Different versions that address minor issues MUST NOT change the device’s sample data format or alter an existing register map. However, register additions that do not affect the existing register map are allowed. Any change that warrants a modification of a device’s streaming data format, read size, write size, or existing register map MUST be implemented as a new [Device ID](https://open-ephys.github.io/ONI/v1.0/hw-spec/devices.html#dev-id). 

The v2 change corresponds a commit wherein HubClock data bytes are re-ordered which changes the "sample data format". This change (changing format and incrementing v1 -> v2) occurred in other devices as well (e.g. TS4231V1 and BNO055 devices). In this unique circumstance, we decided to just decrement the minimum version required. Any problems that users might have with this should be documented in bonsai-onix1-docs.

cjsha:
~In this PR, I assumed this belongs to the same release cycle as 0.8.0 instead of having its own 0.7.1.~

Edit, I force-pushed revision 0.7.1

Edit, fix #618

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -13,7 +13,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <PackageIcon>icon.png</PackageIcon>
-    <VersionPrefix>0.7.0</VersionPrefix>
+    <VersionPrefix>0.7.1</VersionPrefix>
     <VersionSuffix></VersionSuffix>
     <LangVersion>10.0</LangVersion>
     <Features>strict</Features>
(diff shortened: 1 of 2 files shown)


bparks13 (approved):
Approved for v0.7.1

PR #621, opened 2026-04-13
Move towards generic headstage dialog

jonnew:
- Generic headstage configuration GUI to which device specific tabs are added programtatially
- Reduces boilderplate and need to create WYSIWYG template for every headstage
- To add another headstage, use the following template:

```csharp
public class ConfigureHeadstageMyNewDialog : HeadstageDialog
{
    internal NeuropixelsV2eDialog DialogProbeA => (NeuropixelsV2eDialog)GetProbeDialog(0);
    internal readonly GenericDeviceDialog DialogBno055;
    internal readonly Headstage64OpticalStimulatorSequenceDialog OpticalStimulatorSequenceDialog;

    public ConfigureHeadstageMyNewDialog(ConfigureHeadstageMyNew configureHeadstage)
    {
        Text = "MyNew Headstage Configuration";

        const string probeName = nameof(ConfigureHeadstageMyNew.ProbeA);
        AddProbeTab(probeName, new NeuropixelsV2eDialog(configureHeadstage.ProbeA, probeName, true),
            old => RecreateDialog(old));

        DialogBno055 = new GenericDeviceDialog(configureHeadstage.Bno055, true);
        AddDeviceTab("Bno055", DialogBno055);

        OpticalStimulatorSequenceDialog = new(configureNode.OpticalStimulator, true);
        AddSequenceTab("Optical Stimulator", OpticalStimulatorSequenceDialog);
    }

    static NeuropixelsV2eDialog RecreateDialog(NeuropixelsV2eDialog old) { ... }
}
```


OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
```diff
@@ -93,6 +93,9 @@ void ButtonOk_Click(object sender, EventArgs e)
             }
         }
 
+        bool IStimulusSequenceDialog.CanCloseForm(out DialogResult result, string stimulusName)
```
bparks13, line 96: Is there a benefit to this pattern instead of marking `CanCloseForm` as public so it satisfies the interface requirement?

bparks13: Edit: this explicit implementation of an interface method allows it to be a private method, but still accessible when cast to the interface type explicitly, keeping the public API clear while still enforcing the contract for internal validation.

OpenEphys.Onix1.Design/Headstage64Dialog.cs
```diff
@@ -1,12 +1,11 @@
...
 namespace OpenEphys.Onix1.Design
 {
     /// <summary>
-    /// Partial class to create a channel configuration GUI for a <see cref="Headstage64Dialog"/>.
+    /// GUI for <see cref="ConfigureHeadstage64"/>. Hosts two stimulus sequence dialogs
+    /// (Electrical and Optical Stimulator) and three <see cref="GenericDeviceDialog"/> tabs
+    /// (Rhd2164, Bno055, TS4231), each in its own tab.
     /// </summary>
-    public partial class Headstage64Dialog : Form
+    internal class Headstage64Dialog : HeadstageDialog
```
bparks13, line 8: Inconsistent accessibility; other headstage dialogs such as `NeuropixelsV1eHeadstageDialog` and `HeadstageRhs2116Dialog` are public classes, whereas `Headstage64Dialog` is now internal.

bparks13: Related: the `Headstage64Editor` is an internal class, while most editors are public classes. In Bonsai, the editors and dialogs appear to be mostly public, however I think it best to move all dialogs and editors to be internal for now unless we plan to expose them for extensibility down the road. This was a mistake on my part when first developing the Design library. There is no good reason to mark them all as public, since the only place they should realistically be called is from the main library.

bparks13 (approved):
Looks good, tested the dialogs extensively to ensure that all functionality is intact. I did find a couple small inconsistencies that I pushed fixes for.

There are some merge conflicts that I can fix (I think it's all XML related), but I wanted to wait and see if any other changes were made to `issue-590` before fixing anything.

I am approving this PR, but I want to discuss changes to the accessibility level for all editors and dialogs as per the comment below. That is outside of the purview of this PR, but I think it is important to be consistent in our implementation. I mistakenly marked some as internal and some as public; since these are all UI-related, I would recommend we mark them all as internal.

OpenEphys.Onix1.Design/GenericStimulusSequenceDialog.cs
jonnew: The reason is the resolved, implicitly access modifier of Interface members is public. Any direct implementation needs to be declared public and affects API surface. Further, any derived class would need to deal with the implementation (e.g. an override would require marking it virtual) rather than just pointing the interface member to something else.

OpenEphys.Onix1.Design/Headstage64Dialog.cs
jonnew: Seems right to me. I don't see the design library getting a lot of use outside this context.

jonnew:
OK, I think ive provided required feedback to finish the merge.

PR #622, opened 2026-04-14
Allow I2C reties for PolledBno055Data

jonnew:
- This commit implements a retry scheme within PolledBno055Data. A register read cycle can be retried 10 times before an exception is propogated.
- This might be a bad idea because it masks underlying communication issues that can result in silent failures on headtages that use the raw deserializer for communication. However, these failures really should be detected on their own since the coincidental failure of a backchannel i2c read is not a robust mechanism for detecting them.


jonnew:
After discussing with @aacuevas, we both agree that this is a hack around something with a proper solution.

PR #624, opened 2026-04-17
Fix error message regarding failed version check

### LINKED ISSUE #623: Change error message for device version/software version incompatibility
The error message currently advises users to update the firmware, but we do not currently provide any way for users to perform this update.

https://github.com/open-ephys/bonsai-onix1/blob/dbc5a24f745ce5c39b808b16b64bdaae8d99cfc0/OpenEphys.Onix1/ContextHelper.cs#L105-L106

cjsha:
We do not provide a way to update the firmware, so any mention of updating firmware should be deleted.

Fix #623


OpenEphys.Onix1/ContextHelper.cs
```diff
@@ -102,8 +102,7 @@ static void ThrowInvalidDeviceVersionException(Type expectedType, uint address,
 
             Console.Error.WriteLine($"Error: The {expectedType.Name} device at address {address} is v{deviceVersion}, " +
                 $"but v{minimumVersion} is required by {libraryName} {libraryVersion}.");
-            Console.Error.WriteLine($"In order to use {libraryName} {libraryVersion} with this device, you will need to update it firmware. " +
-                $"Firmware update files and instructions can be found at https://open-ephys.github.io/onix-docs/index.html.");
+            Console.Error.WriteLine("In order to use this device, you will need to use an older version of the software.");
```
jonnew: Console.Error.WriteLine($"In order to use this device, you will need to use a previous version of {libraryName} that is compatible with {expectedType.Name} v{deviceVersion}.");

jonnew (changes requested):

OpenEphys.Onix1/ContextHelper.cs
jonnew: Also, you'll need to bump the library version to 0.7.2 or 0.8.0.

jonnew:
I accidentlly nuke this things history. Starting over in new branch

Issue #576, opened 2026-02-13
BufferedDataFrame arrays are not guaranteed to have the same length


bparks13:
While working on #247, I added an implicit assumption that the length of the `Clock` array is the number of samples that all other arrays / Mat channels will have; however, this is not true, at least it is not guaranteed to be true. There is nothing explicitly in the class or any of the concrete classes that checks if these arrays are the same length.

Most of the arrays and Mat objects do currently have the same number of samples, but this is not true for the `NeuropixelsV1eDataFrame`, specifically the `LfpData` Mat has 1/12th the number of samples as the `SpikeData` Mat due to a difference in sample rate.

The question then is how can we enforce that the arrays are all the same size at compile time? Not possible with arrays, since these can be dynamically changed at runtime due to the different buffer sizes across data frames

Are there other data structures that could allow this? We could create a custom struct which holds a single clock/hub clock value pair, and we then pass an array of these structs to ensure that we have a 1:1 mapping for clock values, but this would be a breaking change for all current constructors.

What are other options?
- Debug assertions: In debug mode, check the length of the clock/hub clock properties in the constructor
No performance hit for Release builds, but also does not catch any errors in release mode; this should be fine as long as we catch any errors in size during development. Concrete classes would then be in charge of checking if certain arrays are the correct size. 
 - Run-time checking: Every time the constructor is called, validate the two array lengths; this adds an overhead to every constructor call, which is not ideal for performance.
 
We could add debug assertions to enforce size validation during development, which should ensure that all frames are correct in release mode; all other options would add too much code or break the current API surface.

A secondary addition could be adding a `Count` property to the `BufferedDataFrame` abstract class that returns the number of samples; if we have already checked that the `Clock` and `HubClock` properties are the same size in the constructor, then we can return the length of either property as the `Count`, removing the need for the ArrowWriter to pull the length of the array directly.

PR #625, opened 2026-04-24
Add FrameWriter

### LINKED ISSUE #247: Proposal: DataFrameWriter

### LINKED ISSUE #619: Add estimated sample rate attribute to data frames
This is relevant for the frames themselves, since it would be helpful to know the estimated sample rate of a data frame. This is particularly useful for the `FrameWriter`, where we can take this attribute and calculate the optimal buffer size for the data frames directly.

As a first pass, I have created this table with some starting values and notes.

| Device           | Real Sample Rate     | Estimated Sample Rate | Notes                                                                 |
|------------------|----------------------|-----------------------|-----------------------------------------------------------------------|
| Bno055           | 100 Hz nominally     | 100 Hz                | Could be polled at a lower speed                                      |
| DigitalIO        | 10 MHz               | 1000 Hz               | Dependent on how many digital events occur                            |
| HarpSyncInput    | 1 Hz                 | 1 Hz                  | Should we have a minimum buffer size so that this is forced higher?   |
| Heartbeat        | ?? - 10e6 Hz         | 100 Hz                | Minimum rate is device-dependent; configurable by the user            |
| LoadTester       | ??                   | ?                     | Configurable frames per second                                        |
| MemoryMonitor    | 1–1000 Hz            | 100 Hz                | Configurable rate                                                     |
| PortController   | ??                   | 10 Hz                 | Only sends data when state changes                                    |
| Rhs2116Trigger   | ??                   | 1000 Hz               | Dependent on when stimuli are applied                                 |
| TS4231V1         | ??                   | ?                     | Dependent on how many sensors receive a signal                        |

bparks13:
This PR adds a new node (`FrameWriter`) that can be used to save full frames directly to an Apache Arrow file. Each channel in the data frame is saved as its own channel, including nested channels such as `Quaternion` or `Vector3`. Data frames are discovered using reflection, and the schema is built during run-time. Expression trees are compiled to convert data frames to Arrow `RecordBatch`'s. 

I also added a few new attributes [`ExpectedSampleRate`, `FrameWriterIgnore`, and `FrameWriterIgnoreProperties`]. The expected sample rate is also added to relevant data frames where they have an expected sample rate; user-configurable devices do not have an expected sample rate. The sample rate is used to set the buffer size, saving one second's worth of data in memory before writing it to disk. If no attribute is given, it goes to the default buffer size of 1,000. If data is received slowly, there is a timeout every 5 seconds that flushes any data to the file.

## Saving
In terms of performance, it takes slightly longer to write data to the file as compared to the `MatrixWriter`, due to the file format and metadata overhead. To test this, I generated an hour's worth of NeuropixelsV1e data frames and pushed it through the reactive pipeline. 
- For the `FrameWriter`, the whole frame is saved in the *.arrow file directly.
- For the `MatrixWriter`, I extract all of the individual data streams and save them to individual files, since the `MatrixWriter` cannot write more than one stream per file.

| Implementation                         | Samples per Frame | Batch Size | Save Time        | Save Time (seconds) |
|--------------------------------------|-------------------|-------------|------------------|---------------------|
| FrameWriter  | 36 | 30000 | 4 min 20 sec | 260 |
| MatrixWriter                         | 36             | n/a   | 1 min 38 sec     | 98                  |

## Loading
The Arrow file format requires that the data be loaded before it can be handled, which increases the time it takes to load and view data. The data can be memory-mapped so it does not overflow the RAM, but this still requires some file IO compared to creating a memory map of a binary file. There are some optimizations we can explore by integrating a new Extractor in SpikeInterface that would allow us to load the data lazily, decreasing the time it takes to load data.

Loading data using `pyarrow` takes 71.2 ± 8.2 seconds to load the 1-hour long recording as a PyArrow Table. Users can choose arbitrary start/end times within this file without incurring any performance penalties; in fact, choosing a subset of the data to load decreases the time to load proportionally.

Fixes #247 
Fixes #619


OpenEphys.Onix1/FrameWriter/ArrowWriter.cs
```diff
@@ -0,0 +1,65 @@
...
+            writer.WriteRecordBatch(batch);
+        }
+
+        /// <summary>
+        /// Disposes of the resources used by the instance, optionally releasing managed resources if disposing is true.
+        /// </summary>
+        /// <param name="disposing">True to release both managed and unmanaged resources; false to release only unmanaged resources.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (Interlocked.CompareExchange(ref isDisposed, 1, 0) == 0)
```
jonnew: Although a reasonable mechanism to guard this call, it does not make sense in the broader context of this class. For instance, `ArrowWriter` is not thread safe. This guards against multiple threads calling `Dispose` but does nothing for a race between `Write` and `Dispose`.  Replace with standard boolean pattern  because this gives a false sense of thread safety while not providing any.

OpenEphys.Onix1/DataFrameWriter/ArrowWriter.cs
```diff
@@ -0,0 +1,65 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Releases the resources used by the instance.
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
```
jonnew, line 66: Add `GC.SuppressFinalize(this);` This is the standard `IDisposable` pattern. Without it, the finalizer (if the runtime ever adds one) would run unnecessarily.

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            return member switch
+            {
+                FieldInfo fieldInfo => fieldInfo.GetValue(instance),
+                PropertyInfo propertyInfo => propertyInfo.GetValue(instance),
+                _ => throw new InvalidOperationException($"Cannot get value of {member.GetType()} member from {instance.GetType()} object."),
+            };
+        }
+
+        static Schema GenerateSchema(IEnumerable<MemberInfo> members, object instance)
```
jonnew: Move to Helper

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// Represents an operator that writes each data frame in the sequence
+    /// to an Apache Arrow file using an <see cref="ArrowWriter"/>.
+    /// </summary>
+    [WorkflowElementCategory(ElementCategory.Sink)]
+    public class FrameWriter : FileSink
+    {
+        readonly TimeSpan BufferTimeout = TimeSpan.FromSeconds(5);
+        const int DefaultBufferSize = 1000;
+        const int MinimumBufferSize = 50;
+        const int MaximumBufferSize = 10000;
```
jonnew: This seems arbitrary. What if the source is a camera with megapixel images? What is the reason for adding this?

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
```diff
@@ -0,0 +1,217 @@
...
+        /// </returns>
+        public IObservable<BufferedDataFrame> Process(IObservable<BufferedDataFrame> source)
+        {
+            Schema schema = null;
+            Func<IList<BufferedDataFrame>, Schema, RecordBatch> createRecordBatch = null;
+            int bufferSize = DefaultBufferSize;
+
+            return source.Replay(frames => Observable.Concat(
+                frames.Take(1)
+                    .Do(frame =>
```
jonnew: Although this gets the job done, its an abuse of reactive operators. What you have done here is found a way to reintroduce stateful, non-functional programming into a functional paradigm.

`Do` is for observation. Its side effects should have produce some change in "global" world in some way. Write to file, visualize data, affect hardware. In this case, Do is abused to affect the only most local of state, only visible to the capture of the lambda in `Defer`. Given that state's locality, it should be obvious that it should be eliminated and replaced with the _return value_ of some function if we are thinking in functional context. 

The way to do this with Rx is `Select` to transform a frame into a schema etc, followed by a `SelectMany` to map that configuration to the entire future sequence.

OpenEphys.Onix1/FrameWriter/ArrowWriter.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                if (disposing)
+                {
+                    writer.WriteEnd();
```
jonnew: What happens if this call throws?

OpenEphys.Onix1/FrameWriter/ArrowWriter.cs
```diff
@@ -0,0 +1,65 @@
...
+            stream = new FileStream(filename, FileMode.Create);
+            writer = new(stream, schema);
+            writer.WriteStart();
+        }
+
+        /// <summary>
+        /// Writes the specified record batch to the underlying stream.
+        /// </summary>
+        /// <param name="batch">The record batch to write.</param>
+        public void Write(RecordBatch batch)
```
jonnew: This should not be public. The class is public and the member is public and therefore this is now part of the 
API surface (a huge decision). its used precisely once by a derived class. 

Additionally, stylistically, you could use a function alias for this.

OpenEphys.Onix1/FrameWriter/ArrowBatchWriter.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        /// <param name="filename">The path to the output file.</param>
+        /// <param name="schema">The schema describing the structure of the data.</param>
+        /// <param name="bufferSize">The maximum number of items to buffer before writing a batch.</param>
+        /// <param name="timeout">The maximum time to wait before writing a batch.</param>
+        /// <param name="createRecordBatch">A delegate to create a RecordBatch from a list of items and a schema.</param>
+        public ArrowBatchWriter(string filename, Schema schema, int bufferSize, TimeSpan timeout, Func<IList<T>, Schema, RecordBatch> createRecordBatch)
+            : base(filename, schema)
+        {
+            this.schema = schema;
```
jonnew: You are assigning external state that is captured later in a function when you could be using function arguments and no exposed state.

OpenEphys.Onix1/FrameWriter/ArrowBatchWriter.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        /// <param name="item">The item to add to the buffer.</param>
+        public void Write(T item)
+        {
+            subject.OnNext(item);
+        }
+
+        void WriteBatch(IList<T> items)
+        {
+            var recordBatch = createRecordBatch(items, schema);
```
jonnew: How did schema and createRecordBatch get in here?  They came from the ether.

OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameExpressionProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                            Expression.ArrayAccess(arrowArrays, arrowArrayIndex),
+                            Expression.Call(
+                                convertMatRowMethod,
+                                combinedMat,
+                                rowIndex,
+                                matElementType,
+                                batchRows)),
+                        Expression.PostIncrementAssign(arrowArrayIndex));
+
+                    var forLoop = Expression.Block(
```
jonnew, line 101: I'm not good at this expression tree stuff, but I don't think there is any runtime check for the size of a `Mat` matching that required by the file schema. This is a difference compared to `BinaryWriter` which will just write the binary block to the file and not care about size. 

If a later frame following the one used to create the schema has a `Mat`  with a different row count than the first, the schema and the actual arrays will be mismatched (different column count), producing a corrupt Arrow file or a runtime exception with no clear message. A guard at write time comparing `mat.Rows` against the schema column count would catch this early.

Maybe one way around this is our own enforcement about Frames. Maybe concrete Frame implementations (e.g. Rhd2000DataFrame in the #591 allows both 16 channel bipolar and 32 channel unipolar to use the same Frame type and maybe this should not be allowed or the channel size should be made explicit like having fixed sizes as part of their constructors or, more strictly, as par of their type).

OpenEphys.Onix1/FrameWriter/BufferedDataFrameExpressionProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                arrowBuffer = new ArrowBuffer(rowManager.Memory);
+                nullBitmap = ArrowBuffer.Empty;
+            }
+            else
+            {
+                byte[] buffer = new byte[batchRows * mat.ElementSize];
+
+                fixed (byte* bufferPtr = buffer)
+                {
+                    StrideCopy<ushort>(rowManager.Memory.Pin().Pointer, bufferPtr, length, stride);
```
jonnew: This function is meant to be generic over the `Depth` of elements within the `Mat`, but `T` is hardcoded to `ushort`. So if the `Mat` contained something other than an 2-byte element this would result in a segfault or data corruption

jonnew: `rowManager.Memory.Pin()` returns a `MemoryHandle` that implements `IDisposble` and therefore should be wrapped in `using`.

OpenEphys.Onix1/DataFrameWriter/DataFrameWriterHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+                Depth.U16 => GetArrowType(typeof(ushort)),
+                Depth.S16 => GetArrowType(typeof(short)),
+                Depth.S32 => GetArrowType(typeof(int)),
+                Depth.F32 => GetArrowType(typeof(float)),
+                Depth.F64 => GetArrowType(typeof(double)),
+                _ => throw new NotSupportedException($"Cannot get the ArrowType for the given depth value '{depth}'.")
+            };
+        }
+
+        internal static IEnumerable<MemberInfo> GetDataMembers(Type type)
```
jonnew, line 95: If this is on a hot path, or called redundantly enough, then consider a local `Dictionary<Type, IEnumerable<MemberInfo>>` cache.

OpenEphys.Onix1/FrameWriter/BufferedDataFrameExpressionProvider.cs
jonnew: I think this path only gets activated for cases like the LFP data in which there are Mats of different size in the buffered data frame, which is why you probably were not catching this in your test fixtures. This indicates the test fixtures are not exhaustive (which is kind of expected since writing exhaustive is hard, but you should try!). Really they currently very lightly taxing the library instead of poking it in the eye which is what you should be doing.

OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameExpressionProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+    class BufferedDataFrameExpressionProvider : IRecordBatchExpressionProvider
+    {
+        public ParameterExpression InputParameter { get; }
+
+        public BufferedDataFrameExpressionProvider()
+        {
+            InputParameter = Expression.Parameter(typeof(IList<BufferedDataFrame>), "frame");
+        }
+
+        public Expression GetLengthExpression()
```
jonnew, line 21: This expression amounts to

`frames[0].Clock.Length * frames.Count`

- What if frames is empty? This is guarded against in practice using the  `.Where(list => list.Count > 0)` but, given that this function is public member of internal class, it needs to be robust on its on its own merits. Currently, the protection against this is invisible and it would need to always be used in a similar way to prevent failure.
- Is this a good assumption to make, in general? Does it point to the need for frames to have a fixed length as a part of their type (or as close to that as C# allows)?

OpenEphys.Onix1/FrameWriter/ArrowBatchWriter.cs
```diff
@@ -0,0 +1,76 @@
...
+                .Buffer(timeout, bufferSize)
+                .Where(list => list.Count > 0)
+                .Subscribe(WriteBatch);
+        }
+
+        /// <summary>
+        /// Adds an item to the buffer and flushes the buffer when it reaches its maximum size.
+        /// </summary>
+        /// <param name="item">The item to add to the buffer.</param>
+        public void Write(T item)
```
jonnew: `base.Write` exists but this is not an override, confusing.

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        static int ClampBufferSize(int bufferSize) => Math.Min(Math.Max(bufferSize, MinimumBufferSize), MaximumBufferSize);
+
+        static int GetBufferSize(Type frameType)
+        {
+            var sampleRateAttribute = frameType.GetCustomAttribute<ExpectedSampleRateAttribute>();
+            if (sampleRateAttribute != null)
+            {
+                const double BufferDurationSeconds = 1.0;
+                int bufferSize = (int)(sampleRateAttribute.SampleRateHz * BufferDurationSeconds);
```
jonnew: I know that we decided this is the way to go, but the fact that there is no information about the size of a Frame taken into account feels wrong. What if the frame is sampled at 1 Hz and each from is 7 GB of data? What if the frame is sampled at 100 MHz, something feels fishy.

jonnew (changes requested):
NOTE: See my PR to your fork, which corrects most of these issues: https://github.com/bparks13/bonsai-onix1/pull/18

Overall, this is good and I appreciate the effort. There are some loose ends and potential improvements that I've committed (please review my changes), but this is on the right track. 

One thing that is not correct yet is the Test Fixtures. For the ones that produce "real" data frames, e.g. Bno055DataFrame etc, you need to implement in a way that does not require modifying the library via adding constructors etc. That means simply emulating hardware (make the byte[] array and map using the Payload structs). However, this is minor compared to `FrameWriterTesting`, which is correct in theory but no good implementation because its not a test. The point of any Test is to be mean as hell to the library. e.g.

```
        public class FrameWriterTestingBufferedFrame : BufferedDataFrame
        {
            static byte[] bytes = new byte[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 };

            public Mat MatProp { get; } = Mat.FromArray(bytes, 3, 4, Depth.U8, 1);

            public int[] IntProp { get; } = new int[] { 42, 42, 42, 42 };

            public double[] DoubleProp { get; } = new double[] { 3.14, 3.14, 3.14, 3.14 };

            public FrameWriterTestingBufferedFrame(ulong[] clock, ulong[] hubClock) : base(clock, hubClock)
            {
            }
        }
```

OK, but what happens with MatProp has 1e6 elements? What happens when it a different Depth? What happens when its Z dimension is not 1 (remember it actually allows 4 elements per X, Y index)? If there are failures, then maybe they are fixable here or maybe our `DataFrame` needs a better idea of what it really is. The main point is that this is not a real test because its not stressing the library.

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
bparks13: I reviewed the changes made in your PR, and I think we can simplify it even further. I have added a new [commit](https://github.com/bparks13/bonsai-onix1/pull/18/changes/72a41c88c84c5c04726e64b347570a5e18eeb623) that moves the processing of the frame into the `CreateWriter` method, since we already have access to the first frame in that method but we weren't using it in any way. This allows us to remove the Replay/Select/SelectMany pattern, while maintaining the original functionality.

OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameExpressionProvider.cs
bparks13: I created an issue a few months ago (#576) discussing something close to this issue.

When I started working on the frame writer we talked about this particular issue, and the conclusion at the time was to restrict the Arrow writer to only operate on data frames from the Onix1 library, where we ensure that the Mat objects do not change the number of channels (`mat.Rows`) in the Mat from frame to frame. This means that it is not a general solution for Mat objects, as they have no such restrictions in the larger context, but we made that concession specifically so that we didn't have to check every incoming frame for it's size.

That being said, if there are scenarios where the Mat could change size during acquisition, then we absolutely do need some run-time checks for the total number of columns in a frame. It's not hard to add, but would increase the performance cost for every frame as it is running in the hot path. Let me know if I should work on this, or if the implicit assumption for Mat's can be expected to hold for now.

OpenEphys.Onix1/FrameWriter/BufferedDataFrameExpressionProvider.cs
bparks13: I think this also points to the larger issue of needing unit tests for the library as a whole, which we do not have. I just made a [new issue](https://github.com/open-ephys/bonsai-onix1/issues/628) to track this, as it is an important topic.

OpenEphys.Onix1/DataFrameWriter/DataFrameWriterHelper.cs
bparks13: This method is only called during the initial construction of the expression tree, and is not in the hot path.

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
bparks13: If the frame is sampled at 1 Hz we have the timeout now that will cap it at 5 seconds.

More generally, since we have access to the `Schema` we can estimate the size in MB of each frame; the question is then how would we best use this information? During testing, a larger buffer size always led to better performance regardless of the size of the frame (`DigitalIO` and `NeuropixelsV1e` both saw orders of magnitude faster write/read times with increasing buffer sizes). If we then want to have a cap on memory (for very large frames), we might be reducing overall performance. Now, if the frame is actually 7 GB, that is probably a worthy tradeoff so we don't hold 700 GB in memory, but that is always something we can add in later if we develop a frame that requires it.

Let me know if this is something we should work on for this PR, or create an issue to keep track of it.

OpenEphys.Onix1/DataFrameWriter/DataFrameWriterHelper.cs
```diff
@@ -11,6 +11,9 @@ namespace OpenEphys.Onix1.DataFrameWriter
 {
     static class DataFrameWriterHelper
     {
+        const int DefaultBufferSize = 1000;
```
jonnew, line 18: why did you remove my documentation?

OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
```diff
@@ -17,87 +11,8 @@ namespace OpenEphys.Onix1.DataFrameWriter
     [WorkflowElementCategory(ElementCategory.Sink)]
     public class DataFrameWriter : FileSink
     {
-        /// <summary>
```
jonnew: why did you remove my documentation? I know that its private, i still want to be able look at the code a know what this highly important parameter is.

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
jonnew: Yes, this is much better. Nice. 

When reviewing these I noticed something: `FileSink` exposes `Buffered` property. This is actually very interesting though if you look at its documentation. Does this actually work? I dont see a reason why it shouldn't. I think this means that using FileSink gives us the ability to use separate thread for free. 

This line shows how this works: https://github.com/bonsai-rx/bonsai/blob/848b4298b7e67ad5d920f147a918b1dae40a6bd9/Bonsai.System/IO/WriterDisposable.cs#L15

OpenEphys.Onix1/FrameWriter/FrameWriter.cs
jonnew: yeah, i think this might be a case were the magnitude of the corner case actually matters and 7 GB at 1 Hz is so far away from a real data source that its meaningful.

100 MHz would probably be heavily buffered and would have a small packet size, and critically, would probably be given an `ExpectedSampleRate`.

I guess it comes down to this implicit reality that, unless we attaching this thing to some data pipe going into amazon data center, 1-5 second's worth of data can only be so much and even in the extreme cases it will just result in large batch sizes, which is actually good.

I guess the only thing that is really not covered is a high-bandwidth data source that does not have an expected sample rate property. In that case, the default of 1000 samples/batch will be very low performance. I wonder if we should just make this number quite high and rely on the timeout to take care of low bandwidth sources since they don't need to be efficient anyway (they don't produce much data and that's always the easiest to deal with). .

jonnew (approved):
See my comments

OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameExpressionProvider.cs
bparks13: Created a new issue #630 to discuss the size issue. During the technical meeting, we decided to move forward with the current architecture for now, and we will revisit it when this issue is resolved.

PR #627, opened 2026-04-29
Use backing field in LinearTransform3D's A and B setters

### LINKED ISSUE #626: Use backing field in LinearTransform3D's A and B setters
`UpdateM` uses the properties as inputs. Its cleaner (esp. since this class is not sealed) to use the backing field. Replace

https://github.com/open-ephys/bonsai-onix1/blob/64296afa69230ba768b06737e976d748f26e58da/OpenEphys.Onix1/LinearTransform3D.cs#L20

with

```
public Matrix4x4 A { get => a; set { a = value; M = UpdateM(a, B); } }
```

jonnew:
- Fixes #626

### DIFF
--- a/OpenEphys.Onix1/LinearTransform3D.cs
+++ b/OpenEphys.Onix1/LinearTransform3D.cs
@@ -18,13 +18,13 @@ namespace OpenEphys.Onix1
         /// The A matrix in A * <see cref="M"/> = <see cref="B"/>. It is
         /// constructed from a set of four pre-transform Cartesian coordinates.
         /// </summary>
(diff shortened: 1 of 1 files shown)


bparks13 (commented):
We should update the call so that both backing fields are used when calling `UpdateM`. This commit only changes the A or B property that is being updated to the backing field, but based on the initial issue's point (that this is not a sealed class) we should be consistent with where we use the property vs. the backing field.

jonnew:
Im closing this because the correct solution, which is to remove `[LinearTransform3D` entirely, and only expose a Matrix4x4 in  the linear transform class leads to usability issues. This backing field thing is debatable at best.

PR #642, opened 2026-05-11
Add writer task

### LINKED ISSUE #632: ObjectDisposedException can be thrown from output nodes
This was initially discovered when testing for issue #455.

There is a race condition when shutting down the workflow where the `ContextTask` can be disposed before `DigitalOutput` or `AnalogOutput` nodes are finished writing data. 

```mermaid
flowchart TD
    A[Start acquisition with large read size] -->|Streaming data to output nodes| B[Press stop acquisition]
    B --> |Tasks finish| C[ContextTask disposed]
    C -->|Output nodes receive one more set of values| D[Attempts to write using ContextTask]
    D --> E[Throw ObjectDisposedException]
```

Here are the lines that are throwing the exception when called by either output node after disposal:
https://github.com/open-ephys/bonsai-onix1/blob/b87e44cae5de0fc68b19adce88cdd13fe96cf31a/OpenEphys.Onix1/ContextTask.cs#L556-L570

### Potential Solution

To avoid this race condition, we could follow the same pattern as `EnsureContext`, where we check `disposed` before writing:
https://github.com/open-ephys/bonsai-onix1/blob/b87e44cae5de0fc68b19adce88cdd13fe96cf31a/OpenEphys.Onix1/ContextTask.cs#L521-L530

### Workflow
Here is a complete workflow that is consistently triggering this exception:

<details>
  <summary>
  Workflow with Breakout Board, DigitalOutput, and AnalogOutput nodes to trigger an ObjectDisposedException.
  </summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<WorkflowBuilder Version="2.9.0"
                 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                 xmlns:onix1="clr-namespace:OpenEphys.Onix1;assembly=OpenEphys.Onix1"
                 xmlns:rx="clr-namespace:Bonsai.Reactive;assembly=Bonsai.Core"
                 xmlns:dsp="clr-namespace:Bonsai.Dsp;assembly=Bonsai.Dsp"
                 xmlns="https://bonsai-rx.org/2018/workflow">
  <Workflow>
    <Nodes>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="onix1:CreateContext">
          <onix1:Driver>riffa</onix1:Driver>
          <onix1:Index>0</onix1:Index>
        </Combinator>
      </Expression>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="onix1:ConfigureBreakoutBoard">
          <onix1:Name>BreakoutBoard</onix1:Name>
          <onix1:Heartbeat>
            <onix1:DeviceName>BreakoutBoard/Heartbeat</onix1:DeviceName>
            <onix1:DeviceAddress>0</onix1:DeviceAddress>
            <onix1:BeatsPerSecond>100</onix1:BeatsPerSecond>
          </onix1:Heartbeat>
          <onix1:AnalogIO>
            <onix1:DeviceName>BreakoutBoard/AnalogIO</onix1:DeviceName>
            <onix1:DeviceAddress>6</onix1:DeviceAddress>
            <onix1:Enable>false</onix1:Enable>
            <onix1:InputRange0>TenVolts</onix1:InputRange0>
            <onix1:InputRange1>TenVolts</onix1:InputRange1>
            <onix1:InputRange2>TenVolts</onix1:InputRange2>
            <onix1:InputRange3>TenVolts</onix1:InputRange3>
            <onix1:InputRange4>TenVolts</onix1:InputRange4>
            <onix1:InputRange5>TenVolts</onix1:InputRange5>
            <onix1:InputRange6>TenVolts</onix1:InputRange6>
            <onix1:InputRange7>TenVolts</onix1:InputRange7>
            <onix1:InputRange8>TenVolts</onix1:InputRange8>
            <onix1:InputRange9>TenVolts</onix1:InputRange9>
            <onix1:InputRange10>TenVolts</onix1:InputRange10>
            <onix1:InputRange11>TenVolts</onix1:InputRange11>
            <onix1:Direction0>Output</onix1:Direction0>
            <onix1:Direction1>Input</onix1:Direction1>
            <onix1:Direction2>Input</onix1:Direction2>
            <onix1:Direction3>Input</onix1:Direction3>
            <onix1:Direction4>Input</onix1:Direction4>
            <onix1:Direction5>Input</onix1:Direction5>
            <onix1:Direction6>Input</onix1:Direction6>
            <onix1:Direction7>Input</onix1:Direction7>
            <onix1:Direction8>Input</onix1:Direction8>
            <onix1:Direction9>Input</onix1:Direction9>
            <onix1:Direction10>Input</onix1:Direction10>
            <onix1:Direction11>Input</onix1:Direction11>
          </onix1:AnalogIO>
          <onix1:DigitalIO>
            <onix1:DeviceName>BreakoutBoard/DigitalIO</onix1:DeviceName>
            <onix1:DeviceAddress>7</onix1:DeviceAddress>
            <onix1:Enable>false</onix1:Enable>
            <onix1:DeadTime>0</onix1:DeadTime>
            <onix1:SampleRate xsi:nil="true" />
          </onix1:DigitalIO>
          <onix1:ClockOutput>
            <onix1:DeviceName>BreakoutBoard/OutputClock</onix1:DeviceName>
            <onix1:DeviceAddress>5</onix1:DeviceAddress>
            <onix1:ClockGate>false</onix1:ClockGate>
            <onix1:Frequency>1000000</onix1:Frequency>
            <onix1:DutyCycle>50</onix1:DutyCycle>
            <onix1:Delay>0</onix1:Delay>
          </onix1:ClockOutput>
          <onix1:HarpInput>
            <onix1:DeviceName>BreakoutBoard/HarpSyncInput</onix1:DeviceName>
            <onix1:DeviceAddress>12</onix1:DeviceAddress>
            <onix1:Enable>false</onix1:Enable>
            <onix1:Source>Breakout</onix1:Source>
          </onix1:HarpInput>
          <onix1:MemoryMonitor>
            <onix1:DeviceName>BreakoutBoard/MemoryMonitor</onix1:DeviceName>
            <onix1:DeviceAddress>10</onix1:DeviceAddress>
            <onix1:Enable>false</onix1:Enable>
            <onix1:SamplesPerSecond>10</onix1:SamplesPerSecond>
          </onix1:MemoryMonitor>
        </Combinator>
      </Expression>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="onix1:StartAcquisition">
          <onix1:ReadSize>128</onix1:ReadSize>
          <onix1:WriteSize>2048</onix1:WriteSize>
        </Combinator>
      </Expression>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="rx:Timer">
          <rx:DueTime>PT0S</rx:DueTime>
          <rx:Period>PT0.01S</rx:Period>
        </Combinator>
      </Expression>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="IntProperty">
          <Value>1</Value>
        </Combinator>
      </Expression>
      <Expression xsi:type="rx:Accumulate" />
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="onix1:DigitalOutput">
          <onix1:DeviceName>BreakoutBoard/DigitalIO</onix1:DeviceName>
        </Combinator>
      </Expression>
      <Expression xsi:type="GroupWorkflow">
        <Name>Ramp Generator</Name>
        <Workflow>
          <Nodes>
            <Expression xsi:type="Combinator">
              <Combinator xsi:type="rx:Timer">
                <rx:DueTime>PT0S</rx:DueTime>
                <rx:Period>PT0.01S</rx:Period>
              </Combinator>
            </Expression>
            <Expression xsi:type="Combinator">
              <Combinator xsi:type="FloatProperty">
                <Value>0.2</Value>
              </Combinator>
            </Expression>
            <Expression xsi:type="rx:Accumulate" />
            <Expression xsi:type="Mod">
              <Operand xsi:type="FloatProperty">
                <Value>20</Value>
              </Operand>
            </Expression>
            <Expression xsi:type="Add">
              <Operand xsi:type="FloatProperty">
                <Value>-10</Value>
              </Operand>
            </Expression>
            <Expression xsi:type="Combinator">
              <Combinator xsi:type="rx:BufferCount">
                <rx:Count>12</rx:Count>
                <rx:Skip>1</rx:Skip>
              </Combinator>
            </Expression>
            <Expression xsi:type="dsp:ConvertToArray">
              <dsp:Depth xsi:nil="true" />
            </Expression>
            <Expression xsi:type="WorkflowOutput" />
          </Nodes>
          <Edges>
            <Edge From="0" To="1" Label="Source1" />
            <Edge From="1" To="2" Label="Source1" />
            <Edge From="2" To="3" Label="Source1" />
            <Edge From="3" To="4" Label="Source1" />
            <Edge From="4" To="5" Label="Source1" />
            <Edge From="5" To="6" Label="Source1" />
            <Edge From="6" To="7" Label="Source1" />
          </Edges>
        </Workflow>
      </Expression>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="onix1:AnalogOutput">
          <onix1:DeviceName>BreakoutBoard/AnalogIO</onix1:DeviceName>
          <onix1:DataType>Volts</onix1:DataType>
        </Combinator>
      </Expression>
      <Expression xsi:type="Combinator">
        <Combinator xsi:type="onix1:HeartbeatData">
          <onix1:DeviceName>BreakoutBoard/Heartbeat</onix1:DeviceName>
        </Combinator>
      </Expression>
      <Expression xsi:type="MemberSelector">
        <Selector>Clock</Selector>
      </Expression>
    </Nodes>
    <Edges>
      <Edge From="0" To="1" Label="Source1" />
      <Edge From="1" To="2" Label="Source1" />
      <Edge From="3" To="4" Label="Source1" />
      <Edge From="4" To="5" Label="Source1" />
      <Edge From="5" To="6" Label="Source1" />
      <Edge From="7" To="8" Label="Source1" />
      <Edge From="9" To="10" Label="Source1" />
    </Edges>
  </Workflow>
</WorkflowBuilder>
```
</details>

bparks13:
Similar to the existing infrastructure for `readFrames` and `distributeFrames`, I added a `Task` that handles `writeFrames`. 

This uses a `Channel`, specifically an unbounded multiple-producer-single-consumer channel, which operates as a queue (first-in-first-out [FIFO]).

All write method calls now add an action to the queue, which is triggered based on events using an `async/await` pattern to minimize wasteful CPU cycles.

Fixes #632

### DIFF
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -462,14 +462,9 @@ namespace OpenEphys.Onix1
                     {
                         try
                         {
-                            while (!collectFramesToken.IsCancellationRequested)
+                            await foreach (var writeAction in writeQueue.Reader.ReadAllAsync(collectFramesToken))
                             {
-                                var writeAction = await writeQueue.Reader.ReadAsync(collectFramesToken);
-
-                                lock (writeLock)
-                                {
-                                    writeAction();
-                                }
+                                writeAction();
                             }
                         }
                         catch (OperationCanceledException)
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -501,7 +501,6 @@ namespace OpenEphys.Onix1
                             }
                             frameQueue?.Dispose();
                             frameQueue = null;
-                            while (writeQueue?.Reader.TryRead(out _) == true) { }
                             writeQueue = null;
                             ctx.Stop();
 
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -6,6 +6,7 @@ using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Threading.Channels;
 
 namespace OpenEphys.Onix1
 {
@@ -55,7 +56,9 @@ namespace OpenEphys.Onix1
         bool disposed;
         Task readFrames;
         Task distributeFrames;
+        Task writeFrames;
         Task acquisition = Task.CompletedTask;
+        Channel<Action> writeQueue;
         CancellationTokenSource collectFramesCancellation;
         event Func<ContextTask, IDisposable> ConfigureAndLatchControllerEvent;
         event Func<ContextTask, IDisposable> ConfigureAndLatchLinkEvent;
@@ -291,7 +294,7 @@ namespace OpenEphys.Onix1
             }
             finally
             {
-                 Reset();
+                Reset();
             }
         }
 
@@ -449,7 +452,40 @@ namespace OpenEphys.Onix1
                     TaskCreationOptions.LongRunning,
                     TaskScheduler.Default);
 
-                    return acquisition = Task.WhenAll(distributeFrames, readFrames).ContinueWith(task =>
+                    var options = new UnboundedChannelOptions()
+                    {
+                        SingleReader = true
+                    };
+                    writeQueue = Channel.CreateUnbounded<Action>(options);
+
+                    writeFrames = Task.Run(async () =>
+                    {
+                        try
+                        {
+                            while (!collectFramesToken.IsCancellationRequested)
+                            {
+                                var writeAction = await writeQueue.Reader.ReadAsync(collectFramesToken);
+
+                                lock (writeLock)
+                                {
+                                    writeAction();
+                                }
+                            }
+                        }
+                        catch (OperationCanceledException)
+                        {
+#if DEBUG
+                            Console.WriteLine("Frame write task has been cancelled by " + this.GetType());
+#endif
+                        }
+                        finally
+                        {
+                            writeQueue.Writer.Complete();
+                        }
+                    },
+                    collectFramesToken);
+
+                    return acquisition = Task.WhenAll(distributeFrames, readFrames, writeFrames).ContinueWith(task =>
                     {
                         if (readFrames.IsFaulted && readFrames.Exception is AggregateException ex)
                         {
@@ -470,6 +506,8 @@ namespace OpenEphys.Onix1
                             }
                             frameQueue?.Dispose();
                             frameQueue = null;
+                            while (writeQueue?.Reader.TryRead(out _) == true) { }
+                            writeQueue = null;
                             ctx.Stop();
 
                             contextConfiguration.Dispose();
@@ -555,26 +593,17 @@ namespace OpenEphys.Onix1
 
         internal void Write<T>(uint deviceAddress, T data) where T : unmanaged
         {
-            lock (writeLock)
-            {
-                ctx.Write(deviceAddress, data);
-            }
+            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data));
         }
 
         internal void Write<T>(uint deviceAddress, T[] data) where T : unmanaged
         {
-            lock (writeLock)
-            {
-                ctx.Write(deviceAddress, data);
-            }
+            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data));
         }
 
         internal void Write(uint deviceAddress, IntPtr data, int dataSize)
         {
-            lock (writeLock)
-            {
-                ctx.Write(deviceAddress, data, dataSize);
-            }
+            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data, dataSize));
         }
 
         internal oni.Hub GetHub(uint deviceAddress) => ctx.GetHub(deviceAddress);
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -58,7 +58,7 @@ namespace OpenEphys.Onix1
         Task distributeFrames;
         Task writeFrames;
         Task acquisition = Task.CompletedTask;
-        Channel<Action> writeQueue;
+        Channel<WriterTaskAction> writeActionChannel;
         CancellationTokenSource collectFramesCancellation;
         event Func<ContextTask, IDisposable> ConfigureAndLatchControllerEvent;
         event Func<ContextTask, IDisposable> ConfigureAndLatchLinkEvent;
@@ -456,15 +456,15 @@ namespace OpenEphys.Onix1
                     {
                         SingleReader = true
                     };
-                    writeQueue = Channel.CreateUnbounded<Action>(options);
+                    writeActionChannel = Channel.CreateUnbounded<WriterTaskAction>(options);
 
                     writeFrames = Task.Run(async () =>
                     {
                         try
                         {
-                            await foreach (var writeAction in writeQueue.Reader.ReadAllAsync(collectFramesToken))
+                            await foreach (var writeAction in writeActionChannel.Reader.ReadAllAsync(collectFramesToken))
                             {
-                                writeAction();
+                                writeAction.Write(ctx);
                             }
                         }
                         catch (OperationCanceledException)
@@ -475,7 +475,7 @@ namespace OpenEphys.Onix1
                         }
                         finally
                         {
-                            writeQueue.Writer.Complete();
+                            writeActionChannel.Writer.Complete();
                         }
                     },
                     collectFramesToken);
@@ -501,7 +501,7 @@ namespace OpenEphys.Onix1
                             }
                             frameQueue?.Dispose();
                             frameQueue = null;
-                            writeQueue = null;
+                            writeActionChannel = null;
                             ctx.Stop();
 
                             contextConfiguration.Dispose();
@@ -587,12 +587,12 @@ namespace OpenEphys.Onix1
 
         internal void Write<T>(uint deviceAddress, T data) where T : unmanaged
         {
-            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data));
+            writeActionChannel?.Writer.TryWrite(new ScalarWriterTaskAction<T>(deviceAddress, data));
         }
 
         internal void Write<T>(uint deviceAddress, T[] data) where T : unmanaged
         {
-            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data));
+            writeActionChannel?.Writer.TryWrite(new ArrayWriterTaskAction<T>(deviceAddress, data));
         }
 
         internal oni.Hub GetHub(uint deviceAddress) => ctx.GetHub(deviceAddress);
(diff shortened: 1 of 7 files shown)


OpenEphys.Onix1/ContextTask.cs
```diff
@@ -449,7 +452,40 @@ internal Task StartAsync(int blockReadSize, int blockWriteSize, CancellationToke
...
+                    {
+                        SingleReader = true
+                    };
+                    writeQueue = Channel.CreateUnbounded<Action>(options);
+
+                    writeFrames = Task.Run(async () =>
+                    {
+                        try
+                        {
+                            while (!collectFramesToken.IsCancellationRequested)
```
aacuevas: An alternative to consider would be
`await foreach (var writeAction in writeQueue.Reader.ReadAllAsync(collectFramesToken)`

it's less verbose, but both options work

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -470,6 +506,8 @@ internal Task StartAsync(int blockReadSize, int blockWriteSize, CancellationToke
                             }
                             frameQueue?.Dispose();
                             frameQueue = null;
+                            while (writeQueue?.Reader.TryRead(out _) == true) { }
```
aacuevas: Is this dump necessary? the `writeQueue = null` line will remove the queue reference and, in turn, this will delete all inner objects, right?

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -555,26 +593,17 @@ private oni.Frame ReadFrame()
...
+            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data));
         }
 
         internal void Write(uint deviceAddress, IntPtr data, int dataSize)
         {
-            lock (writeLock)
-            {
-                ctx.Write(deviceAddress, data, dataSize);
-            }
+            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data, dataSize));
```
aacuevas: This is extremely dangerous.

An `IntPtr` does not hold data by itself. It is, as its name implies, a pointer to somewhere else.

Before, this was a blocking call, so it was safe to handle the pointer. However now this queues an action into another task. By the time this action gets executed, nothing guarantees the underlying data still exists. (e.g, if this comes from a `Mat`, the `Mat` might have been released)

aacuevas (changes requested):
While the overall idea is good, there is a very dangerous case that needs to be addressed.

OpenEphys.Onix1/ContextTask.cs
bparks13: I agree, I will put the less verbose option in.

OpenEphys.Onix1/ContextTask.cs
bparks13: `writeQueue = null` removes the reference, so eventually GC will run and take care of all of the objects, but I thought it would be better to be explicit about draining the queue. Directly above this, we follow the same pattern of draining to clear the `frameQueue`, which I know is different because those are `frame`'s that need to be disposed of, but the overall pattern seemed valid to me.

OpenEphys.Onix1/ContextTask.cs
bparks13: @aacuevas that is an excellent point, I was only focused on the two `Write` methods that I saw throwing the exception, and neglected to test this path.

What would be the best way forward? As far as I know, there is nothing we can do to pin the underlying data in this method, because all we have is the `IntPtr` itself.

Should we copy the data to an array and pass it to the array overload for `Write`? This would add an extra copy in the hot path for writes, but it sidesteps the potential issue of the data being missing when the queue processes this item.

bparks13: As per our discussion earlier, this overload has been removed. The only node that used it, `AnalogOutput`, has been modified to call the `Write(T[])` overload instead, with the data copied into an array that is then passed to the method instead of an `IntPtr`.

bparks13:
@aacuevas I have modified the code as we discussed. I am also opening a new PR targeting this branch so we can discuss other patterns for how to handle the writer queue so that we don't pollute the commit history here.

bparks13:
@aacuevas The WriterTaskAction PR has been merged here, and this PR is now ready for review again.

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -449,7 +452,35 @@ internal Task StartAsync(int blockReadSize, int blockWriteSize, CancellationToke
                     TaskCreationOptions.LongRunning,
                     TaskScheduler.Default);
 
-                    return acquisition = Task.WhenAll(distributeFrames, readFrames).ContinueWith(task =>
+                    var options = new UnboundedChannelOptions()
```
aacuevas: call this `channelOptions` or something similar. even if it's local, it's always best to be clear with variable naming

aacuevas (changes requested):
There is one major issue with the implementation. If there are any exceptions on the write calls, the writer task will exit but the `WaitAll` will still wait for the other two tasks to end.

At the very least, an exception on that task should trigger the cancellation token, same as the read task does. This will stop the workflow, but at this point an exception on write means some hardware issue, so it is ok.

The only issue is that, the way it is now, bonsai will not show the exception as originating from the calling node, but probably from the write task itself. Changing this might entail a more extensive change of the `ContextTask` API, though.

bparks13:
@aacuevas That's true that the exception stack trace will look like it was the `ContextTask` that threw because of the writer queue. We do have access to the device address that threw the exception (currently it's a private member in `WriterTaskAction`, but we could expose that in some way). This way we could throw a new exception that includes the device address (or get the device from `DeviceTable` and display the device string?) with the original exception as the inner exception to preserve the full error message.

aacuevas:
> @aacuevas That's true that the exception stack trace will look like it was the `ContextTask` that threw because of the writer queue. We do have access to the device address that threw the exception (currently it's a private member in `WriterTaskAction`, but we could expose that in some way). This way we could throw a new exception that includes the device address (or get the device from `DeviceTable` and display the device string?) with the original exception as the inner exception to preserve the full error message.

I don't think any of this is necessary.
Having the data-emitting node appear as the source of the exception could be good for user perception. But the truth is, any Write exceptions are independent of the device and truly originate at the context level, so the context task being marked as the origin is not wrong.

But we need to catch those exceptions and handle the error consequently still.

bparks13:
During testing of this PR, we discovered issue #658. This PR is marked as a draft until that issue has been resolved so we can fully test that the writeTask being added is being handled correctly during exceptional circumstances.

aacuevas:
Given that this is intended to fix an old bug and seems to be intertwined with another old bug, but it's a corner case unlikely to happen in most cases, let's move it for next release but focus on it as soon ad the current release is out.

PR #643, opened 2026-05-12
Add compression option for DataFrameWriter

### LINKED ISSUE #639: Add compression functionality to Arrow file writers
As of May 6th, `Apache.Arrow` has been updated to v23.0.0, as well as the `Apache.Arrow.Compression` package. The compression package update includes updates to transitive dependencies such as `ZstdSharp`, which brings the `System.Memory` dependency up to the same version the rest of `Apache.Arrow` uses, which is also what Bonsai uses. This now allows us to bring in the compression package and enable compression, whereas previously it would throw run-time exceptions due to assembly name mismatches.

This update should include a new property that the user can toggle, which is then propagated to the `FileWriter`. This does not require any changes in the loading scripts, as packages such as `pyarrow` automatically detect if compression is enabled and handle the conversion behind the scenes.

bparks13:
Updated `Apache.Arrow` to v23.0.0, and added a dependency on `Apache.Arrow.Compression` which allows us to compress the data frames before they are written to the file.

While adding the compression option, I realized that I did not correctly configure the inheritance structure, so the `DataFrameWriter` did not have the same contract as `DataFrameFileSink` and `BufferedDataFrameFileSink`. To ensure that all relevant classes contained the property, and could be propagated down the hierarchy correctly, I had to add the abstract `ArrowFileSink` class, and attempted to make sure that everything was correctly lined up. Jon had to simplify the structure for me, removing the unnecessary interface that I added. Now the `EnableCompression` property exists at all levels, and can be correctly pushed through the hierarchy.

Fixes #639

### DIFF
--- a/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
@@ -1,0 +1,9 @@
+﻿using Bonsai.IO;
+
+namespace OpenEphys.Onix1.DataFrameWriter
+{
+    abstract class ArrowFileSink<T1, T2> : FileSink<T1, T2>, IArrowSinkOptions where T2 : ArrowWriter
+    {
+        public bool EnableCompression { get; set; } = false;
+    }
+}
--- a/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
@@ -2,7 +2,7 @@
 
 namespace OpenEphys.Onix1.DataFrameWriter
 {
-    abstract class ArrowFileSink<T1, T2> : FileSink<T1, T2>, IArrowSinkOptions where T2 : ArrowWriter
+    abstract class ArrowFileSink<TSource> : FileSink<TSource, ArrowBatchWriter<TSource>>
     {
         public bool EnableCompression { get; set; } = false;
     }
--- a/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
@@ -9,7 +9,7 @@ namespace OpenEphys.Onix1.DataFrameWriter
     /// to an Apache Arrow file using an <see cref="ArrowWriter"/>.
     /// </summary>
     [WorkflowElementCategory(ElementCategory.Sink)]
-    public class DataFrameWriter : FileSink, IArrowSinkOptions
+    public class DataFrameWriter : FileSink
     {
         const int SecondsBeforeFlush = 5;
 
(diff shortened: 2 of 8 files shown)


OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Bonsai.IO;
+
+namespace OpenEphys.Onix1.DataFrameWriter
+{
+    abstract class ArrowFileSink<T1, T2> : FileSink<T1, T2>, IArrowSinkOptions where T2 : ArrowWriter
```
jonnew: In our case, `TWriter` is fully determined by `TSource` (since it's always `ArrowBatchWriter<TSource>`), so this leaks an implementation detail into the type signature unnecessarily.

Side note: why remove the descriptive generic names `Tsource` and `Twriter` with `T1` and `T2`?

OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
```diff
@@ -42,13 +41,17 @@ public IObservable<BufferedDataFrame> Process(IObservable<BufferedDataFrame> sou
...
         public IObservable<DataFrame> Process(IObservable<DataFrame> source)
         {
-            return new DataFrameArrowFileSink(TimeSpan.FromSeconds(SecondsBeforeFlush))
-            {
-                FileName = this.FileName,
-                Suffix = this.Suffix,
-                Buffered = this.Buffered,
-                Overwrite = this.Overwrite
-            }.Process(source);
+            return ConfigureSink(new DataFrameArrowFileSink(TimeSpan.FromSeconds(SecondsBeforeFlush))).Process(source);
```
jonnew: I think this function obfuscates for little reason. The previous object initializers were very easy to understand.

jonnew (commented):
Looking at this briefly. Most of these changes are not actually related to the #639. I'm actually a bit unsure what problem they are solving and that should be stated in the commit message.

- IMO, a lot of these changes make the code harder to understand
- A whole interface was added so that `ConfigureSink` could be used, but I don't agree that this function adds clarity or even reduction in lines of code. 
- If you are using generics, try to use descriptive names for generics unless its completely obvious from the implementation what they are for (e.g. some simple math operation). Especially true when inheriting from an generic type, don't re-direct the type parameters, just use the ones from upstream. e.g. don't rename `Tsource` and `Tsink` generics `T1` and `T2`.

image

- The abstract `ArrowFileSink` is ok, but all it was doing before was adding a property for another step of inheritance. In my change, the `TWriter` is hard coded. If you really want to be able to use a `TWriter` other than `ArrowBatchWriter` down the line, you could readd the two-param version of `ArrowFileSink`.

OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
bparks13: I was quickly iterating over this, and wasn't thinking about it when I added `T1` / `T2`. Thank you for simplifying the inheritance structure, it makes a lot more sense now.

OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
bparks13: My main thought with adding the function is to have one place where we add any new parameters from the `ArrowFileSink` class, instead of needing to add it in two places. Other than that, I agree that it doesn't add any benefit.

jonnew (approved):

PR #646, opened 2026-05-13
Create WriteAction class

bparks13:
In #642, @aacuevas brought up a good point that the infrastructure there might be too generic in that it accepts any arbitrary `Action` with no constraints.

This PR attempts to constrain the write queue to only allow `WriteAction` objects instead of generic Actions. The `WriteAction` class is a private class that holds the data to be written, and only executes a `ctx.Write()` call using a passed context, so no additional state is held other than the data to write and the device address.

If this is a valid pattern, we can merge it into the existing PR and then merge to main to keep the commit history clean, otherwise we can close this branch.

### DIFF
--- a/OpenEphys.Onix1/ContextTask.cs
+++ b/OpenEphys.Onix1/ContextTask.cs
@@ -34,6 +34,42 @@ namespace OpenEphys.Onix1
     /// </remarks>
     public class ContextTask : IDisposable
     {
+        abstract class WriteAction
+        {
+            readonly uint deviceAddress;
+
+            internal abstract void Write(oni.Context ctx);
+
+            public WriteAction(uint deviceAddress)
+            {
+                this.deviceAddress = deviceAddress;
+            }
+
+            internal sealed class Scalar<T> : WriteAction where T : unmanaged
+            {
+                readonly T data;
+
+                public Scalar(uint deviceAddress, T data) : base(deviceAddress)
+                {
+                    this.data = data;
+                }
+
+                internal override void Write(oni.Context ctx) => ctx.Write(deviceAddress, data);
+            }
+
+            internal sealed class Array<T> : WriteAction where T : unmanaged
+            {
+                readonly T[] data;
+
+                public Array(uint deviceAddress, T[] data) : base(deviceAddress)
+                {
+                    this.data = data;
+                }
+
+                internal override void Write(oni.Context ctx) => ctx.Write(deviceAddress, data);
+            }
+        }
+
         readonly oni.Context ctx;
 
         /// <summary>
@@ -58,7 +94,7 @@ namespace OpenEphys.Onix1
         Task distributeFrames;
         Task writeFrames;
         Task acquisition = Task.CompletedTask;
-        Channel<Action> writeQueue;
+        Channel<WriteAction> writeQueue;
         CancellationTokenSource collectFramesCancellation;
         event Func<ContextTask, IDisposable> ConfigureAndLatchControllerEvent;
         event Func<ContextTask, IDisposable> ConfigureAndLatchLinkEvent;
@@ -456,7 +492,7 @@ namespace OpenEphys.Onix1
                     {
                         SingleReader = true
                     };
-                    writeQueue = Channel.CreateUnbounded<Action>(options);
+                    writeQueue = Channel.CreateUnbounded<WriteAction>(options);
 
                     writeFrames = Task.Run(async () =>
                     {
@@ -464,7 +500,7 @@ namespace OpenEphys.Onix1
                         {
                             await foreach (var writeAction in writeQueue.Reader.ReadAllAsync(collectFramesToken))
                             {
-                                writeAction();
+                                writeAction.Write(ctx);
                             }
                         }
                         catch (OperationCanceledException)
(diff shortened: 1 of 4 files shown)


aacuevas:
Seems way clearer to me, more constrained and goes from an anti-pattern (direct delegate queuing) to something akin to a command pattern, with clear responsibilities.

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -34,6 +34,42 @@ namespace OpenEphys.Onix1
     /// </remarks>
     public class ContextTask : IDisposable
     {
+        abstract class WriteAction
```
jonnew: If you are going to go this route, then move this somewhere else in the file. There are a bunch of important constants and state right after this that deserve to be near the top of the file.

This class is also highly nested. There are two inner levels of class definitions is a smell for sure. The derived classes being nested inside their base class is very smelly because it inverts the normal mental model where a base class shouldn't "know about" or contain its derivatives.

Overall, use access modifiers to restrict the scope of a class instead of embedding into this class unless you need to access private state from ContextTask.

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -34,6 +34,42 @@ namespace OpenEphys.Onix1
...
+
+                public Scalar(uint deviceAddress, T data) : base(deviceAddress)
+                {
+                    this.data = data;
+                }
+
+                internal override void Write(oni.Context ctx) => ctx.Write(deviceAddress, data);
+            }
+
+            internal sealed class Array<T> : WriteAction where T : unmanaged
```
jonnew: An Array filled with Ts is actually a `WriteAction`?

Things like Array and Scalar are almost certainly going to shadow other widely used type. Further, these names just don't make sense because they don't describe the class.

How about `ArrayWriteAction`?

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -34,6 +34,42 @@ namespace OpenEphys.Onix1
...
+            public WriteAction(uint deviceAddress)
+            {
+                this.deviceAddress = deviceAddress;
+            }
+
+            internal sealed class Scalar<T> : WriteAction where T : unmanaged
+            {
+                readonly T data;
+
+                public Scalar(uint deviceAddress, T data) : base(deviceAddress)
```
jonnew: See comment for Array<T>

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -58,7 +94,7 @@ public class ContextTask : IDisposable
         Task distributeFrames;
         Task writeFrames;
         Task acquisition = Task.CompletedTask;
-        Channel<Action> writeQueue;
+        Channel<WriteAction> writeQueue;
```
jonnew: writeActionChannel.

OpenEphys.Onix1/ContextTask.cs
```diff
@@ -587,12 +624,12 @@ private oni.Frame ReadFrame()
 
         internal void Write<T>(uint deviceAddress, T data) where T : unmanaged
         {
-            writeQueue?.Writer.TryWrite(() => ctx.Write(deviceAddress, data));
+            writeQueue?.Writer.TryWrite(new WriteAction.Scalar<T>(deviceAddress, data));
```
jonnew: WriteAction.Scalar -> ScalarWriteAction<T> once you take care of nesting issue.

jonnew (changes requested):

OpenEphys.Onix1/ContextTask.cs
aacuevas: I'm not sure I completely agree with this being that smelly. It depends a bit on intent.
This class, the way I understand it, it's not supposed to be a normal class that anyone could inherit and use.

First, regarding the nesting of the derived classes, it is not uncommon when the intention is to emulate a union-style structure (Discriminated union in terms of functional programming). For this to be complete, the abstract class also needs a private empty constructor. This disallows inheritance from anything that is not private to the class. It is a way of making sort of an "abstract sealed" class. It's a class with a fixed set of children that can not be expanded by other files. So it depends if we want to enforce this contract or not for this particular class. If we don't, then we should definitely split it, as it is the only argument in favor of the nesting.

Second, and perhaps less relevant, it is supposed to be just an interim class to transport an intention from the `ContextTask.Write` methods to the Write Task. It is, in fact, overly specific to it. There are arguments about nested classes to avoid namespace pollution for simple, concise and implementation-detail classes. Since it's intended to be used only here and never outside, I believe it might warrant to be nested inside contexttask, but with the current it might be too verbose.

To be honest, we are crashing again against limitations of the .NET and C# versions. For example newer versions have a `file` access modifier which solves the second point quite elegantly (it would not be part of `ContextTask` but not pollute the namespace). 

With all that said, if we want to keep this simple and clean, since we do not require to enforce complex contracts for an internal class, I would:
- Put the class in its own file
- Rename it to `WriterTaskAction`. `WriteAction` is too generic and could clash with anything in the future. Since we're adding the class to the namespace, keep it specific.
- Put the derived classes in their own files. Even if it's only two lines, it's the way we're managing classes in the project and the recommended way in C#.

aacuevas (approved):

PR #647, opened 2026-05-14
Correct Pulse Period to Inter-Pulse Interval

### LINKED ISSUE #603: Fix disparity between property names and GUI labels
Image

bparks13:
In the Headstage64 Electrical Stimulator GUI, a property was mistakenly labeled as "Pulse Period" when it should have been "Inter-Pulse Interval". Renamed the property, and also fixed the stimulus waveform generation in the GUI to correctly reflect the property.

The GUI now correctly plots the stimulus with the inter-pulse interval set, which now should match the physical waveform generated by the device.

image

@cjsha Can you confirm that the device now outputs what is plotted by the GUI?

Fixes #603


cjsha (approved):
in the few different stimulus patterns that I tested, there's perfect parity in waveform in the plot and the waveform measured in the oscilloscope.

PR #653, opened 2026-05-18
Add comments to SecondsBeforeFlush and BufferDurationSeconds variables

### LINKED ISSUE #650: Important constants are buried in DataFrameWriterHelper
The record batch size for those data source with a defined sample rate is defined here:

https://github.com/open-ephys/bonsai-onix1/blob/5ec6ae234c0fcebd328bddc592b448d590929f6f/OpenEphys.Onix1/DataFrameWriter/DataFrameWriterHelper.cs#L204

Its generally a great idea to reduce the scope of a variable so this make sense, but its quite important and rhymes with a bunch of other constants that are defined in a lot of different places:

https://github.com/open-ephys/bonsai-onix1/blob/5ec6ae234c0fcebd328bddc592b448d590929f6f/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs#L14

https://github.com/open-ephys/bonsai-onix1/blob/5ec6ae234c0fcebd328bddc592b448d590929f6f/OpenEphys.Onix1/DataFrameWriter/DataFrameWriterHelper.cs#L18

https://github.com/open-ephys/bonsai-onix1/blob/5ec6ae234c0fcebd328bddc592b448d590929f6f/OpenEphys.Onix1/DataFrameWriter/DataFrameWriterHelper.cs#L24

I would prefer to move this to one of those locations and document with XML comment. The `SecondsBeforeFlush` constant should also be documented.

bparks13:
Moved `BufferDurationSeconds` to a common location for constant variables, and added comments to constant variables that did not have them.

Fixes #650

### DIFF
--- a/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
@@ -12,8 +12,7 @@ namespace OpenEphys.Onix1.DataFrameWriter
     public class DataFrameWriter : FileSink
     {
         /// <summary>
-        /// Number of seconds to wait before flushing any data in the buffer to the Arrow file.
-        /// Ensures that devices with sporadic sample rates are still saved to the file periodically.
+        /// Ensures that data produced by devices with non-uniform or slow sample rates are written to file periodically.
         /// </summary>
         const int SecondsBeforeFlush = 5;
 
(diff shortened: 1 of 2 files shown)


OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
```diff
@@ -11,6 +11,10 @@ namespace OpenEphys.Onix1.DataFrameWriter
     [WorkflowElementCategory(ElementCategory.Sink)]
     public class DataFrameWriter : FileSink
     {
+        /// <summary>
+        /// Number of seconds to wait before flushing any data in the buffer to the Arrow file.
+        /// Ensures that devices with sporadic sample rates are still saved to the file periodically.
```
jonnew: Ensures that data produced by devices with non-uniform or  slow sample rates are written to file periodically.

jonnew (approved):
Small change to text. requested.

PR #656, opened 2026-05-19
Add remarks describing how Arrow compression works

### LINKED ISSUE #649: Explain what type of compression is used in DataFrameWriter
This boolean turns on or off compression

https://github.com/open-ephys/bonsai-onix1/blob/5ec6ae234c0fcebd328bddc592b448d590929f6f/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs#L17

It would be good to explain what algorithm is used and upon which data it act in the \<remarks\> blocks. e.g. are each recordBatch compressed independently?

bparks13:
Specifies the compression type (Zstd) and that it operates on individual record batches.

Fixes #649

### DIFF
--- a/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
@@ -17,9 +17,9 @@ namespace OpenEphys.Onix1.DataFrameWriter
         /// Gets or sets a value indicating whether to enable compression when writing to the Arrow file.
         /// </summary>
         /// <remarks>
-        /// When enabled, the writer will use Zstandard (Zstd) compression to compress the record batches
-        /// before writing them to the file. This can significantly reduce the file size, especially for 
-        /// channels with redundant data. Each record batch will be compressed independently.
+        /// When enabled, record batches are compressed with Zstandard (Zstd) before writing.
+        /// This reduces file size at the cost of additional CPU overhead on read and write. 
+        /// Each record batch is compressed independently.
         /// </remarks>
         public bool EnableCompression { get; set; } = false;
 
--- a/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
@@ -20,7 +20,7 @@ namespace OpenEphys.Onix1.DataFrameWriter
         /// When enabled, data is compressed using the <see
         /// href="https://github.com/facebook/zstd">Zstandard</see> compression algorithm before
         /// writing. This reduces file size at the cost of additional CPU overhead on read and
-        /// write. Each block of data is compressed independently.
+        /// write.
         /// </remarks>
         public bool EnableCompression { get; set; } = false;
 
--- a/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
@@ -17,9 +17,10 @@ namespace OpenEphys.Onix1.DataFrameWriter
         /// Gets or sets a value indicating whether to enable compression when writing to the Arrow file.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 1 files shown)


OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
```diff
@@ -16,6 +16,11 @@ public class DataFrameWriter : FileSink
         /// <summary>
         /// Gets or sets a value indicating whether to enable compression when writing to the Arrow file.
         /// </summary>
+        /// <remarks>
+        /// When enabled, the writer will use Zstandard (Zstd) compression to compress the record batches
+        /// before writing them to the file. This can significantly reduce the file size, especially for 
+        /// channels with redundant data. Each record batch will be compressed independently.
+        /// </remarks>
```
cjsha, line 24: I mostly have a gripe with the part about redundant data. I think technically, within the context of compression, "redundant" might be correct, but I don't know if neuroscientists would interpret it with such context. Also, I see this statement as kinda like saying "this knife cuts especially well on cuttable things". If I were to revise, it would look something like:

```
/// <remarks>
/// When enabled, record batches are compressed with Zstandard (Zstd) before writing.
/// This reduces file size at the cost of additional CPU overhead on read and write. 
/// Each record batch is compressed independently.
/// </remarks>
```

Except I'd keep the future tense if that's what's consistent with the rest of the docs.

cjsha (commented):
I left some feedback, but I'll let someone else have final say on approval

OpenEphys.Onix1/DataFrameWriter/DataFrameWriter.cs
aacuevas: I agree with @cjsha comment.
Specially removing intensity words like "significantly", since they do really depend on the nature of the data. Cris suggestion is agnostic, true (it will be reduced, to some degree) and easy to understand.

jonnew: When enabled, record batches are compressed with Zstandard (Zstd) before writing.

=>

When enabled, record batches are compressed with \<see  href="https://github.com/facebook/zstd">Zstandard</see\> before writing.

- since Zstd is redundant and its also `zstd` if we were going to use it.
- I'd never heard of this so i think we should provide a link to learn what it is.

jonnew: Also, there is a bigger issue I want to mention and this pertains to all documentation and public interfaces for this data frame writer:

Try not to leak implementation details unless required. In this case: what is a "record batch"? Its not mentioned anywhere else in this top-level (from user's perspective) documentation. They have no way of knowing what this means per se, but need to correlate these docs with other things they read in the tutorial etc. 

For example, in my latest commit, I almost called an attribute `RunEndEncodedArrayAttribute` but then realized that this does not at all describe what the attribute would actually document, but the implementation detail used to when this attribute is present. So, i went with `SubsampledAttribute`.

This is a bit more nuanced since this is an important feature of both the Arrow file format and how this compression operates, but I still think we need to maneuver around using this type of jargon unless required.

jonnew: What is a block of data?

cjsha: I'm not sure "record batch" is an implementation detail that needs to be abstracted from the user. It might be a detail that is fundamentally important for the user to understand when it comes to loading segments of data. I'm guessing "block of data" came from bparks13 trying to hide "record batch" from the user as you suggested. Maybe there's a less jargonny way to describe it like in the example you gave about the attribute.

Anyway, sorry, don't mean to add to much noise here. There might be too many cooks in the kitchen reviewing this PR...

aacuevas: Personally. I would leave this as it is now.
A full explanation of how arrow stores the files and how it applies compression could be interesting as its own page on the docs, but this xmlcomment is documenting the property, should not go too much further.

aacuevas (approved):

PR #657, opened 2026-05-19
Refactor DataFrameWriter and encode subsampled data as REE arrays

### LINKED ISSUE #651: Arrow Run End encoded columns for for NeuropixelV1 LFP Data
I noticed this Arrow column-encoding type:

https://arrow.apache.org/docs/format/Intro.html#run-end-encoded-layout

which specializes at representing runs of repeating values. Given that the LFP data in NeuropixelsV1 frames is mostly null because its sample rate is 1/12 that of spiking data, is it possible to encode these columns with that encoding type, e.g. via some attribute?

### LINKED ISSUE #654: Redundant reflection within DataFrameWriter
Arrow schema generation and expression generation are independent reflection passes on incoming data types. `DataFrameWriterHelper.GenerateSchema` and `IRecordBatchExpressionProvider.GetArrayPopulationExpressions` both independently walk the same member tree via reflection to decide how to encode each member. This means:

1. The same structural decisions (is this a `Mat`? is it sub-sampled?) are made twice, in two places that could diverge.
2. The schema, which already encods what each output Arrow array should look like, is not consulted by the expression builder. 

Expression generation should use the schema to determine the appropriate transform on incoming data to convert to Arrow columns.

jonnew:
- Replace naive Arrow array encoding for sub-sampled BufferedDataFrame members with run-end encoded (REE) arrays (available since Apache.Arrow 23.0.0).
- ForNeuropixelsV1 LFP data:
    - Absolute best case, e.g. using separate file for LFP: 1.0x file size
    - REE: 1.15x file size (with all run ends pre-calculated, zero runtime computational overhead)
    - Null-mask array encoding: 1.85x file size (with large bitmask runtime computational overhead)
- Introduce new [Subsampled(divisor: N)] attribute, where N is the number of primary-rate samples per sub-sampled value. LfpData is decorated with [Subsampled(divisor: 12)].
- DataFrameWriter internals are restructured to eliminate redundant reflection traversals.
- Introduce MemberFieldGroup, pairing each leaf MemberNode with its Arrow schema Field(s). Mats with N rows produce N fields; all other members produce one, making the schema the single source of truth for how each member encodes.
- Replace GenerateSchema with BuildFieldMappings, which produces MemberFieldGroup entries in a single traversal. BuildSchema derives the schema from those groups directly, so schema generation and expression generation share one member walk instead of two independent reflection passes.
- Update IRecordBatchExpressionProvider.GetArrayPopulationExpressions to accept IEnumerable<MemberFieldGroup>.
- Decompose BufferedDataFrameExpressionProvider.GetArrayPopulationExpressions into three focused methods: BuildArrayExpression, BuildMatExpression, and BuildRunEndEncodedMatExpression.
- Simplify DataFrameExpressionProvider: struct expansion now happens in BuildFieldMappings, so the expression provider iterates only leaf groups.
- Fixes #654 
- Fixes #651

### DIFF
--- a/OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameArrowFileSink.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameArrowFileSink.cs
@@ -17,9 +17,10 @@ namespace OpenEphys.Onix1.DataFrameWriter
         {
             var frameType = dataFrame.GetType();
             var members = DataFrameWriterHelper.GetDataMembers(frameType);
-            var schema = DataFrameWriterHelper.GenerateSchema(members, dataFrame);
+            var fieldGroups = DataFrameWriterHelper.BuildFieldMappings(members, dataFrame);
+            var schema = DataFrameWriterHelper.BuildSchema(fieldGroups);
             var createRecordBatch = RecordBatchExpressionFactory.CreateBuilder<Func<IList<BufferedDataFrame>, Schema, RecordBatch>>(
-                new BufferedDataFrameExpressionProvider(), frameType, members).Compile();
+                new BufferedDataFrameExpressionProvider(), frameType, fieldGroups).Compile();
             var bufferSize = (int)Math.Ceiling((double)DataFrameWriterHelper.GetBufferSize(frameType) / dataFrame.Clock.Length);
             return new ArrowBatchWriter<BufferedDataFrame>(filename, schema, bufferSize, timeout, createRecordBatch, EnableCompression);
         }
--- a/OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameExpressionProvider.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameExpressionProvider.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using Apache.Arrow;
@@ -15,7 +14,7 @@ namespace OpenEphys.Onix1.DataFrameWriter
 
         public BufferedDataFrameExpressionProvider()
         {
-            InputParameter = Expression.Parameter(typeof(IList<BufferedDataFrame>), "frame");
+            InputParameter = Expression.Parameter(typeof(IList<BufferedDataFrame>), "bufferedDataFrames");
         }
 
         public Expression GetLengthExpression()
@@ -28,8 +27,8 @@ namespace OpenEphys.Onix1.DataFrameWriter
             var clockArray = Expression.Property(firstFrame, nameof(BufferedDataFrame.Clock));
             var clockArrayLength = Expression.ArrayLength(clockArray);
             var numberOfFrames = Expression.Property(
-                    Expression.Convert(InputParameter, typeof(ICollection<BufferedDataFrame>)),
-                    nameof(ICollection<BufferedDataFrame>.Count));
+                Expression.Convert(InputParameter, typeof(ICollection<BufferedDataFrame>)),
+                nameof(ICollection<BufferedDataFrame>.Count));
             return Expression.Multiply(clockArrayLength, numberOfFrames);
         }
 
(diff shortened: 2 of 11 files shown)


OpenEphys.Onix1/DataFrameWriter/FrameWriterTesting.cs
jonnew, line 1: I didnt mean to add this, please remove cleanly.

bparks13 (commented):
Before I get too far into this, I want to confirm the intention behind this PR, because it's not documented in the notes.

With these changes, when the data is loaded by `pyarrow`, the LFP channels have the same number of valid samples as the AP channels. When the channel is decoded, it looks like this:

| Sample # | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | ... |
| --- |  --- |  --- |  --- |  --- |  --- |  --- |  --- |  --- |  --- | --- | --- | --- | --- | --- |
| Value | N1 | N1 | N1 | N1 | N1 | N1 | N1 | N1 | N1 | N1 | N1 | N1 | N2 |

Where N1 is the first value, N2 is the second, and so on. There is the change in value every 12 samples, but the output does not make it clear that this data is sub-sampled compared to other channels. 

**Question**: Is this what we want the LFP data to look like when it is loaded? 

For context, the original implementation produced data that looked like this without any additional processing:

| Sample # | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | ... |
| --- |  --- |  --- |  --- |  --- |  --- |  --- |  --- |  --- |  --- | --- | --- | --- | --- | --- |
| Value | N1 | null | null | null | null | null | null | null | null | null | null | null | N2 |

To determine what the real LFP values are with the original implementation, it was easy enough to create a mask that could filter for the sub-sampled values by calling:

```python
mask = table['LfpDataCh0'].is_valid()
lfpChannel0 = table['LfpDataCh0'].filter(mask)
clock = table['Clock'].filter(mask)
```

With the new implementation, this changes a little bit and requires detecting changes in the value to know when the LFP timestamp should be pulled from the 'Clock' channel. 

Let me know if this is a valid way to store and load the data, but I feel like it makes it more confusing for the end users because now the data seems to be sampled at 30 kHz with a lot of repeating values. This also becomes more complicated on the processing side; we can add a script specific to Neuropixels V1 data to show how to load it, but someone building a new pipeline from scratch might not understand how to manipulate the data.

bparks13:
As per @aacuevas suggestion, we could add null values to the REE array to match the original output.

This would be similar to the original implementation where the data was stride-copied to an Arrow array and we generated a null bitmap array to indicate where the null values were.

Concretely, with the change I am proposing we would generate the values array and the run_ends array like this:

```
run_ends = [1, 11, 1, 11, 1, 11, ...]
values = [1, null, 2, null, 3, null, ...]
```

To create the values array, we would need to loop through the `Mat` row and interleave null values, which I believe means we cannot make use of any zero-copy procedures. These are not represented as `null` directly in .NET, but rather we would build the null bitmask for the values array.

jonnew:
I'd push back on the characterization that this PR lacks intention: the notes and linked issues make the motivation fairly explicit. The previous implementation was storing data at roughly 2x the necessary size for a device that's already disk-hungry, and this addresses both the space and CPU overhead while improving code quality.

On the clock question, a couple of points worth clarifying:

- A complete LFP round robin spans 12 superframes, each containing a full round robin of spiking data. Saying "this exact time is when the LFP data occurred" is therefore actually not physically correct. Each row is more precisely valid for a specific set of 32 of the 384 channels. 
- That said, I take the point that a single start time per conversion enables phase adjustment in SpikeInterface, which is a real tradeoff worth considering.

If that capability is important, we could accommodate it fairly easily by changing the run end array definition from:
```
[12, 24, 36, ...]
```
to:
```
[1, 12, 23, 24, 35, 36, ...]
```
...and furnishing the data child array with a repeated validity bitmap of:
```
..._01010101_01010101
```
where every other run marks the end of a null block. These are constant per frame writer, so they can be precomputed. There will be a disk space impact, obviously. I'm not sure its worth it.

jonnew:
I think that if you are planning on doing the stride stuff and modifying the mat before its written, then I'm back to disagreeing that this change is worth it. That is extra CPU overhead, diskspace, code, etc for what amounts to an interpretation issue that can be documented.

bparks13:
My apologies, I misspoke in my comment. The initial intent of the issue/PR is quite clear, what I should have said was that the change to how data is loaded was not specifically mentioned, and I wanted to confirm that the data format when loading is acceptable. 

I can add a section to the Neuropixels V1 documentation on how to interpret `DataFrameWriter` files, as well as provide some guidance on how to extract the values from the columns, as it sounds like the data format is acceptable as long as it is documented properly.

aacuevas:
> As per @aacuevas suggestion, we could add null values to the REE array to match the original output.
> 
> This would be similar to the original implementation where the data was stride-copied to an Arrow array and we generated a null bitmap array to indicate where the null values were.
> 
> Concretely, with the change I am proposing we would generate the values array and the run_ends array like this:
> 
> ```
> run_ends = [1, 11, 1, 11, 1, 11, ...]
> values = [1, null, 2, null, 3, null, ...]
> ```
> 
> To create the values array, we would need to loop through the `Mat` row and interleave null values, which I believe means we cannot make use of any zero-copy procedures. These are not represented as `null` directly in .NET, but rather we would build the null bitmask for the values array.

My question was more in the line of if it was possible to add a bitmask buffer to these fields, but it does not seem possible.

So I agree with Jon, the benefits of REE for these values outweighs the interpretation issue.

bparks13 (approved):
Reviewed the changes, everything looks good. Thanks for taking care of the redundancy in the code, this is a cleaner approach.

I also performed some integration testing, this worked with the following devices:
- `AnalogInput`
- `DigitalInput`
- `MemoryMonitor`
- `Heartbeat`
- `NeuropixelsV1e`
- `Bno055`
- `NeuropixelsV2e`

Specifically, the data nodes for these devices were connected to a `DataFrameWriter` and the files were saved to disk. `MemoryMonitor` was checked to confirm that the data was being written to disk faster than it was produced.

All data was subsequently loaded in Python and checked that all fields were present and correct. This specifically includes the Neuropixels V1e LFP data that is now run-end encoded (REE). Plotting the data shows that it is loaded as if the individual samples were saved, but is more efficient now that we are using REE.

bparks13:
@aacuevas The file has been removed and the merge conflicts with main have been resolved. I am leaving this open in case you have any comments or revisions, feel free to merge this when you are ready.

aacuevas:
Checked, reviewed and tested. No further comments, merging.

PR #666, opened 2026-06-29
Add neuropixles scale and CAR operators

### LINKED ISSUE #665: Neuropixels data processing operators
For downstream visualization and real-time processing purposes available in the  `Bonsa.Dsp` and `Bonsai.Ephys` packages:

1. Bandpass filtering
2. Spike detection
3. Waveform visualization

it would be good to have some basic processing operators that are available in other software toolboxes for neuropixels. Namely,

1. Scaling operators that convert ADC codes into SI units
2. Neuropixels-specific common average referencing. For neuropixels, a common average must be maintained for every ADC, each of which is responsible for multiple channels, since individual ADCs have different offsets. This can be read about here: https://open-ephys.github.io/gui-docs/User-Manual/Plugins/Neuropixels-CAR.html

jonnew:
- Scale operators convert amplifier data to uV
- CAR operators perform neuropixels-optimized common average referencing on amplifier data
- Fixes #665


aacuevas:
Wouldn't it be better for these operators to act over the `Neuropixels*DataFrame` instead oif over the mat? The operators are already neuropixel specific after all and might not have complete sense if jut used with an arbitrary Mat after all.

The main benefit would be not having to manually set the `ProbeVersion` property

jonnew:
I though about this a lot and actually originally implemented them this way. However, the big issue is that both of these operators change the underlying type of the Mat from ushort to float, which frankly felt wrong to me. Especially since CAR is extremely lossy. The DataFrame is supposed to be the thing produced by hardware. Thus when you place the DataFrameWriter after it, there is consistency to what is being recorded. If these transforms produced DataFrames, this is broken.

aacuevas:
What about them accepting dataframes and outputting a Mat?
The logic is still sound, I believe. The input is a neuropixel frame, the output is a data array with transformed data, which now can be used to whatever is needed.

The important thing, IMHO, is that the input should be constrained. Connecting a Mat produced by an arbitrary source to the CAR input should not be allowed (a camera lplImage is technically a Mat!)

jonnew:
I guess that's fine so long as the operators are not meant to be useful after any other transform is produced on the data. e.g. what if for some reason somebody wants to filter the data before a CAR (not advisable, but you understand the point). 

I guess the feeling is that any transform can be a stupid choice. For instance, you could threshold all pixels in an image to 0 to turn it black and loose all information in the image. This feels like you should be able to do it even though there is no data processing circumstance where its a good idea. If you have a matrix with the right dimensions, you can run it through this CAR operator and get some result that's probably not very useful. This one feels odd because the transform is _intended_ to be used in a particular circumstance, but from a functional standpoint its kinda the same thing, right?

This is not rhetorical, i don't know what the best option is and I arrived at the current situation as a compromise.

aacuevas:
It would be, if it did not have a property to select between "Which version of neuropixels does the data come from"
If it were a true CAR, for generic processing of a matrix, it would not care where the data comes from.

That property which says "you can connect anything here, but must be from a neuropixels and you must select which one, if you are mistaken here it will ruin the output" does not feel right

It's the fact that we have this property that is so explicit. And, from the code perspective, that the CAR node actually uses this property to call static methods from the `Neuropixelsv1` and `Neuropixelsv2` classes, using a switch

It really says to me "this node is closely linked to the neuropixel output"

jonnew:
yes, thats fair. I could make an annoying argument about how this is actually the same as putting bad parameters for any transform, but I think you are right from a practical perspective.

jonnew:
Perhaps in this case this should be a `ConvertNeuropixelsAmplifierData` or something:

- Always scales uV
- Optionally  performs CAR

this way both of the operations are constrained. We are converting to floats in CAR anyway and it seems very strange to keep the ADC code units at that point.

jonnew:
@aacuevas The latest commit should address your concerns.

aacuevas (approved):
Looks good and well-constrained now

Issue #674, opened 2026-08-03
Neuropixels configuration GUI refactor


jonnew:
This aggregates feature requests and other issues related to completing the Neuropixels configuration GUI refactor. 

## Feature Ideas
*The following are random ideas I'm having while addressing this issue. They can either graduate to sub-issues or be addressed via discussion as the work progress*

1. **Blocked channel actions**. Lets say a blocked channel has really good activity. You want to be able to (1) unblock it and (2) inspect what pinned electrode will be deactivited. Maybe contact info shows the offending contact? Maybe using a brightly colored line to connect them? maybe you are just allowed to unblock?
1.  **Activity Survey**
    - Certain exceptions should not result in retry. e.g. 
    	- Calibration file mismatch
    	-  Probe part number (NPyyyy) does not match the probe interface file's model name (NPxxxx)
    - Perhaps include filter for non-physiological "spikes" e.g. those that don't have a biploar waveform within the window.
    - A "settings changed since last run" warning banner that extends the datetime warning .e.g. if the user switches on bandpass filtering or selects a different bank
1. **Coherent operation mode**  Currently checkboxes and hidden booleans accessed through key shortcuts dictate the display mode and interaction mode. This needs to be made more coherent. 
    - Display state: channels, banks, or activity
    - Interaction state: Channel select, bank select, or ruler
    - The mode and display state are reflected in text at the bottom left with hints for escaping
1. **Font and scale**
    - Change to scalable font
    - Make the control size independent of monitor resolution
1. **Bank labels**
    - Banks are viewable in two ways at time of writing. 
       1. By selecting electrodes and looking at contact info. This only allows you to see the bank following a selection, not as a passive piece of information presented about the probe. 
       2. By looking at the shade of color in bank select mode. This allows you to see bank boundaries, but not the actual bank letter.
    - Because bank information is useful for channel selections, I think it should somehow be made available as part of the probe viewer. Im not exactly sure how implement this. Currently we do very little on top of the geometry data provided in the PI file, and I dont want to stray from this pattern to much. 
    - The OE gui does the following
      <img width="317" height="413" alt="Image" src="https://github.com/user-attachments/assets/2a5d3305-bf44-427e-9e4e-8f5b8426fbca" />
1. Related to #673 : It might be good to have a scan or verify feature that makes sure the probe interface file matches the probe that is installed.

acomrie:
[...]

Issue #662, opened 2026-06-08
Improve default probe interface file name


jonnew:
the default probe interface file name for a neuropixels probe uses no information about the probe:

<img width="384" height="479" alt="Image" src="https://github.com/user-attachments/assets/186bc6d2-9d97-4bb8-ab78-cc1450f9d0e4" />

Use the address in some way so that file names can be distinguished

riffa_0_porta_npx2

or something.

bparks13:
We would only be able to use information that is available at the headstage configuration node, which I don't believe exposes the driver or the index; what we do have access to is the port and the device names. We can make the filename similar to what is shown in the dropdown menus when choosing a specific device (e.g., "porta_NeuropixelsV2A").  We could also use the full unique identifier: "porta_HeadstageNeuropixelsV2e_NeuropixelsV2A". The benefit of these approaches is that they closely follow the naming scheme of the windows and tabs themselves, which should help users discern which configuration corresponds to which device.

PR #675, opened 2026-08-03
Refactor configuration GUI for NeuropixelsV2e

### LINKED ISSUE #474: Add tooltips to Neuropixels contacts
Similar to the OE GUI, and the previous version of this library, add tooltip on hover over the different electrodes so it is easier to see which one it is, and any relevant information.

### LINKED ISSUE #645: Consider adding a neuropixels probe spike rate map for channel selection
I think I already added this as an issue in the Open Ephys GUI, but would still be useful here.
Users coming from SpikeGLX are used to looking at a probe map in which channels are highlighted based on the spike rate in order to make the channel selection based on current activity. The underlying algorithm for sorting spikes is crude but gets the job done, it scans banks for 30 seconds and spits out a heat map overlaid on the probe.

Image

### LINKED ISSUE #664: Probe Interface misuse
## Lack of data locality
Probe interface is a standard that allows the standardized definitions of probe geometry and metadata, including information like the probe contour, electrode positions etc. However, this library is riddled with code like this:

https://github.com/open-ephys/bonsai-onix1/blob/b146c05b1a1863321d4de724cc9deec34c60c9bb/OpenEphys.Onix1/NeuropixelsV2eProbeGroup.cs#L93

https://github.com/open-ephys/bonsai-onix1/blob/b146c05b1a1863321d4de724cc9deec34c60c9bb/OpenEphys.Onix1/NeuropixelsV2eQuadShankProbeGroup.cs#L84

Why are these being recalculated based on magic consts etc when the contour, electrode positions, and electrode shapes are supposed to be fully defined within the ProbeInterface file? Everything that touches probe interface must be reviewed to resolve this type of issue. My current opinion is that this stuff needs to be removed entirely.

## Reliance on optional fields
There is a check for validity within `NeuropixelsV2eQuadShankProbeGroup` specializatoin of `ProbeGroup` (side issue, Im not sure this class should exist) that ensures the `device_channel_indices` is 384 elements long (side note: this maps the the `DeviceID` key, weirdly)

https://github.com/open-ephys/bonsai-onix1/blob/b146c05b1a1863321d4de724cc9deec34c60c9bb/OpenEphys.Onix1/NeuropixelsV2eQuadShankProbeGroup.cs#L193

However, `device_channel_indices` need not exist in a valid probe interface file so this is wrong.

### LINKED ISSUE #668: Move shared icon resource out of HeadstageDialog
The application icon is currently embedded in `HeadstageDialog.resx` and all other forms that need it borrow it from there via `ComponentResourceManager(typeof(HeadstageDialog))`. The icon belongs to the project, not to that dialog.

The fix is to move the icon into a dedicated static resource class so the dependency is explicit and forms don't need to reference an unrelated dialog to get the company logo.

jonnew:
## Summary
The original goal was to run an electrode activity survey on NeuropixelsV2e probes and persist the results in the probe interface file. That grew into a much larger effort.

Storing per-contact survey results (amplitude, noise, firing rate, and the parameters that produced them) required data OpenEphys.ProbeInterface.NET had no way to carry, due to implementation errors. Fixing that required reworking ProbeInterface.NET's public API, which in turn required migrating every consumer in this repo: probe group classes across NeuropixelsV1/V2 (electrode geometry now comes from the new ProbeInterface.NET model via embedded JSON specs instead of hardcoded Electrode classes), device configs, register contexts, and the WinForms dialogs for NeuropixelsV1, Rhs2116, and others.

The probeinterface re-write landed in the middle of an in-progress NeuropixelsV2 GUI rewrite, so the two efforts ended up combined in one branch rather than split apart, unfortunately.

- NeuropixelsV2e configuration dialog rewritten from WinForms to Dear ImGui (minimap/zoomed probe view, drag-select, contact info, bank-select mode, electrode activity survey UI and persistence).
- Probe geometry sourced from ProbeInterface.NET instead of duplicated into the GUI layer.
- Other ProbeInterface.NET consumers updated for the new API, with incidental improvements where the new API made prior workarounds unnecessary.
- Activity survey implemented along with lots of other features

## Fixes
- Fixes #474
- Fixes #668
- Fixes #664
- Fixes #645

## WIP
- Partially addresses #674 
  - TODO: Survey for NeuropixelsV1e
- Partially addresses #662 
  - TODO: Verify behavior when GUIs have stablized
- #680


jonnew:
For what its worth, this number is misleading because its counting the embedded JSON files as LOC:

image

The count without those is

+5516 / -7,339 

So about 2000 LOC were removed (mostly from resx vs. imgui boilderplate).

PR #677, opened 2026-08-04
Fix DataFrameWriter widening concrete frame types

### LINKED ISSUE #676: DataFrameWriter widens concrete frame types to DataFrame/BufferedDataFrame and violates Sink contract
`DataFrameWriter.cs` is a Bonsai Sink that writes a sequence of data frames to an Apache Arrow file. Two unrelated abstract base classes exist for device frames (DataFrame; BufferedDataFrame), because their payload fields are shaped differently (scalar vs. array/Mat per field). Unifying the hierarchies was considered and rejected.

`DataFrameWriter` has two non-generic Process overloads operating on the abstract base types. Bonsai's binder matches a concrete upstream type (e.g. `NeuropixelsV1DataFrame`) to the `BufferedDataFrame`-typed overload, and the node's output type widens to `IObservable<BufferedDataFrame>` . Downstream nodes lose the concrete subtype's device-specific members. A correct Rx/Bonsai sink must be `IObservable<T> -> IObservable<T>`, identity-preserving.

jonnew:
## Summary
- DataFrameWriter is now a single Bonsai ExpressionBuilder that inspects upstream element type at workflow-build time, picks the matching generic Arrow sink, and emits a call directly against it.
- This preserves concrete frame type with essentially no overhead
- It raises a WorkflowBuildException at build time, so the operator cannot beincorrectly chained during workflow editing
- DataFrameArrowFileSink and BufferedDataFrameArrowFileSink are now generic over the concrete source type instead of fixed to their respective base type, and ArrowFileSink exposes a small IArrowFileSink interface so the builder can configure EnableCompression without knowing the concrete sink type at compile time.
- Fixes #676 

## Note
DataFrameWriter has lost its public `Process` method and thus it has no direct use outside of Bonsai's build system. Direct use in calling code requires using internal classes currently and we might want to remedy this by wrapping it up for use by direct callers:

### Option 1: Use internals
There are no direct callers outside of `Onix1.Design` (#675) which already has access to Onix1 internals. In that case, just use the appropriate `FileSink` directly:

```csharp
IObservable<NeuropixelsV2DataFrame> dataStream = 
    new DataFrameWriter.BufferedDataFrameArrowFileSink<NeuropixelsV2DataFrame>(
        TimeSpan.FromSeconds(DataFrameWriter.DataFrameWriter.SecondsBeforeFlush))
      {
          FileName          = recordingFilePath,
          Suffix            = Bonsai.IO.PathSuffix.None,
          Buffered          = false,
          Overwrite         = true,
          EnableCompression = true
      }.Process(rawStream);
```

### Option 2: Static helpers

```csharp
/// <summary>
/// Writes all of the data frames in the sequence to an Apache Arrow file.
/// </summary>
/// <typeparam name="TSource">The concrete <see cref="DataFrame"/> type in the sequence.</typeparam>
/// <param name="source">The sequence of data frames to write.</param>
/// <param name="fileName">The name of the file on which to write the elements.</param>
/// <param name="suffix">The suffix used to generate file names.</param>
/// <param name="buffered">Indicates whether writing should be buffered.</param>
/// <param name="overwrite">Indicates whether to overwrite the output file if it already exists.</param>
/// <param name="enableCompression">Indicates whether to enable compression when writing to the Arrow file.</param>
/// <returns>
/// An observable sequence identical to <paramref name="source"/>, including its element type, but where
/// there is an additional side effect of writing each frame to an Apache Arrow file.
/// </returns>
public static IObservable<TSource> Write<TSource>(
    IObservable<TSource> source,
    string fileName,
    PathSuffix suffix = PathSuffix.None,
    bool buffered = true,
    bool overwrite = false,
    bool enableCompression = false)
    where TSource : DataFrame
{
    return new DataFrameArrowFileSink<TSource>(TimeSpan.FromSeconds(SecondsBeforeFlush))
    {
        FileName = fileName,
        Suffix = suffix,
        Buffered = buffered,
        Overwrite = overwrite,
        EnableCompression = enableCompression
    }.Process(source);
}

/// <summary>
/// Writes all of the buffered data frames in the sequence to an Apache Arrow file.
/// </summary>
/// <typeparam name="TSource">The concrete <see cref="BufferedDataFrame"/> type in the sequence.</typeparam>
/// <param name="source">The sequence of buffered data frames to write.</param>
/// <param name="fileName">The name of the file on which to write the elements.</param>
/// <param name="suffix">The suffix used to generate file names.</param>
/// <param name="buffered">Indicates whether writing should be buffered.</param>
/// <param name="overwrite">Indicates whether to overwrite the output file if it already exists.</param>
/// <param name="enableCompression">Indicates whether to enable compression when writing to the Arrow file.</param>
/// <returns>
/// An observable sequence identical to <paramref name="source"/>, including its element type, but where
/// there is an additional side effect of writing each frame to an Apache Arrow file.
/// </returns>
public static IObservable<TSource> WriteBuffered<TSource>(
    IObservable<TSource> source,
    string fileName,
    PathSuffix suffix = PathSuffix.None,
    bool buffered = true,
    bool overwrite = false,
    bool enableCompression = false)
    where TSource : BufferedDataFrame
{
    return new BufferedDataFrameArrowFileSink<TSource>(TimeSpan.FromSeconds(SecondsBeforeFlush))
    {
        FileName = fileName,
        Suffix = suffix,
        Buffered = buffered,
        Overwrite = overwrite,
        EnableCompression = enableCompression
    }.Process(source);
}
```

### DIFF
--- a/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
@@ -2,12 +2,7 @@
 
 namespace OpenEphys.Onix1.DataFrameWriter
 {
-    interface IArrowFileSink
-    {
-        bool EnableCompression { get; set; }
-    }
-
-    abstract class ArrowFileSink<TSource, TFrame> : FileSink<TSource, ArrowBatchWriter<TFrame>>, IArrowFileSink
+    abstract class ArrowFileSink<TSource, TFrame> : FileSink<TSource, ArrowBatchWriter<TFrame>>
         where TSource : TFrame
     {
         public bool EnableCompression { get; set; } = false;
--- a/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/ArrowFileSink.cs
@@ -2,7 +2,13 @@
 
 namespace OpenEphys.Onix1.DataFrameWriter
 {
-    abstract class ArrowFileSink<TSource> : FileSink<TSource, ArrowBatchWriter<TSource>>
+    interface IArrowFileSink
+    {
+        bool EnableCompression { get; set; }
+    }
+
+    abstract class ArrowFileSink<TSource, TFrame> : FileSink<TSource, ArrowBatchWriter<TFrame>>, IArrowFileSink
+        where TSource : TFrame
     {
         public bool EnableCompression { get; set; } = false;
     }
--- a/OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameArrowFileSink.cs
+++ b/OpenEphys.Onix1/DataFrameWriter/BufferedDataFrameArrowFileSink.cs
@@ -4,7 +4,8 @@ using Apache.Arrow;
 
 namespace OpenEphys.Onix1.DataFrameWriter
 {
-    class BufferedDataFrameArrowFileSink : ArrowFileSink<BufferedDataFrame>
+    class BufferedDataFrameArrowFileSink<TSource> : ArrowFileSink<TSource, BufferedDataFrame>
+        where TSource : BufferedDataFrame
     {
         readonly TimeSpan timeout;
 
@@ -13,7 +14,7 @@ namespace OpenEphys.Onix1.DataFrameWriter
             this.timeout = timeout;
         }
 
-        protected override ArrowBatchWriter<BufferedDataFrame> CreateWriter(string filename, BufferedDataFrame dataFrame)
+        protected override ArrowBatchWriter<BufferedDataFrame> CreateWriter(string filename, TSource dataFrame)
         {
             var frameType = dataFrame.GetType();
             var members = DataFrameWriterHelper.GetDataMembers(frameType);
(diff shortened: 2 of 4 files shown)


bparks13 (approved):
Looks good, and does correctly fix the error breaking the sink contract. Tested this with `DataFrame`, `BufferedDataFrame`, and unrelated classes; workflow exceptions correctly block unrelated classes, while the concrete types flow through and can be used after the node.

Regarding the open Note: Option 1 seems to already be implemented for the Design library, so no further work needs to be done. Option 2 seems like a good idea to implement here, for two reasons; 1) it closes the public API gap mentioned, and 2) it can simplify the call site in the expression builder. With the proposed static helper, the expression tree simplifies to 

```c#
var instance = Expression.Constant(this);
var writeMethod = typeof(DataFrameWriter)
    .GetMethod(buffered ? nameof(WriteBuffered) : nameof(Write))
    .MakeGenericMethod(elementType);
return Expression.Call(writeMethod, source,
    Expression.Property(instance, nameof(FileName)),
    Expression.Property(instance, nameof(Suffix)),
    Expression.Property(instance, nameof(Buffered)),
    Expression.Property(instance, nameof(Overwrite)),
    Expression.Property(instance, nameof(EnableCompression)));
```

jonnew:
Ok, this is addressed. When merged, we can update #675 to use

```csharp
            IObservable<NeuropixelsV2DataFrame> dataStream = recordingFilePath != null
                ? DataFrameWriter.DataFrameWriter.WriteBuffered(rawStream, recordingFilePath, Bonsai.IO.PathSuffix.None, buffered: false, overwrite: true, enableCompression: true)
                : rawStream;
```     
in `NeuropixelsV2eSurveyRunner.cs`

PR #682, opened 2026-08-11
Fail with exception when NeuropixelsV2 probe metadata can't be read

### LINKED ISSUE #681: ConfigureNeuropixels*PsbDecoder should throw if Enable is true and no probe SN can be read
Currently, in all `ConfigureNeuropixels*PsbDecoders`  if `Enable` is `true` and a probe serial number cannot be read, configuration is skipped, no warning is shown, and no data will ever be produced

https://github.com/open-ephys/bonsai-onix1/blob/d5b83703cc0d0e444073f340d41965b8eaf59044/OpenEphys.Onix1/ConfigureNeuropixelsV2PsbDecoder.cs#L89

thus, probe Flex EEPROM read is a silent enable that is completely opaque to the user. Instead `(Enable && probeMetadata.ProbeSerialNumber is null)` should throw and tell the user that they either need to insert a probe, reseat the probe, or disable the device.

jonnew:
- Remove the `ProbeSerialNumber != null` check that silently skipped NeuropixelsV2/V2Beta configuration whenever the flex EEPROM read failed. The enclosing block now gates on `Enable` alone, so an enabled probe that can't be read now fails configuration instead of streaming nothing with no warning.

- Make `INeuropixelsV2Metadata.ProbeSerialNumber` non-nullable, and have `NeuropixelsV2Metadata`/`NeuropixelsV2eBetaMetadata` always wrap and rethrow `oni.ONIException` as an `InvalidOperationException` instead of swallowing register-read failures (error -5).

- Reword the metadata-read exception across all four probe metadata readers (V1e, V1f, V2, V2eBeta) to name the device and state the two valid remedies: reseat the flex cable, or set Enable to false if the probe is not in use.

- Thread `deviceName` into `NeuropixelsV1eMetadata`, `NeuropixelsV1fMetadata`, and `NeuropixelsV1fRegisterContext` so V1 error messages identify the probe by name instead of I2C address.

- Rename `NeuropixelsV2.ProbeAddress`/`FlexEEPROMAddress` to `ProbeI2CAddress`/`FlexEepromI2CAddress` to match V1 naming.

- Extract magic sleep durations in the V2e/V2 Rhd2000 headstage configuration into named `AnalogSwitchSettleTimeMilliseconds`/ `AnalogSupplySettleTimeMilliseconds` constants. This was required because now disabling a probe speeds up start time consideably and this revealed issues with power supply settling.

- Fixes #681


jonnew:
Superseded by #719

